PR #96, opened 2020-09-04
EOS-12150: HA: JBOD: IO Path Recovery - New decision maker rules files

Madhura-08:
- Call Python API to get server and storage type details
- Logic to load appropriate rules engine file based on server and
  storage_type


ajaysrivas:
This is not required now. @Madhura-08, can we close this PR ?

Madhura-08 (commented):
Yes @ajaysrivas, we don't need this now. So closing the PR.

Madhura-08:
We don't need these changes now. Hence closing the PR.

PR #120, opened 2020-09-17
build-ha-io: support different interfaces on the nodes

andriytk:
## Problem Statement
<pre>
  <code>
    Closes EOS-13356.
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  Not yet.
  </code>
</pre>
## Problem Description
<pre>
  <code>
    Currently, the build-ha-io script expects that the name of the
data network interface is the same on both nodes, which sometimes
may not be the case when the underlying HW is different. (Which
may happen, for example, when the old failed node is replaced
with the new one.)
  </code>
</pre>
## Solution
<pre>
  <code>
   add support for different interface names via the
two new configuration options: --left-iface and --right-iface.
If the options are not explicitly set by user, the value of
the old --iface option is used for both interfaces.
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
    Not yet.
  </code>
</pre>

### DIFF
--- a/conf/script/build-ha-io
+++ b/conf/script/build-ha-io
@@ -46,7 +46,10 @@ Mandatory parameters:
         <CDF>          Hare Cluster Description File
 
 Optional parameters:
-  -i, --interface <if>  Data network interface (default: eth1)
+  -i, --interface <if>  Data network interface (when they are the same
+                        on both nodes, default: eth1)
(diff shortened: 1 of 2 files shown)


andriytk:
Closed in favour of #145.

PR #140, opened 2020-09-23
EOS-13393: fix nodes fencing issue on sporadic hax stop timeouts

andriytk:

   after hax is stopped).

Tested manually on sm10/11-r20.pun setup:
 - build-ha-io PASS
 - failback/failover PASS
 - build-ees-ha-update PASS (with a small timing fix related to #123).

### DIFF
--- a/conf/script/build-cortx-ha
+++ b/conf/script/build-cortx-ha
@@ -27,7 +27,10 @@ init() {
     host2=$(echo $node_list | jq '.["local"][1]' | sed s/\"//g)
 
     # Restart csm to enable decision maker changes
-    pcs resource restart csm-agent
+    cnt=10
+    while ! pcs resource restart csm-agent && ((cnt-- > 0)); do
+        sleep 1
+    done
 
     pcs cluster cib hw_cfg
     # Hardware io stack resource
--- a/conf/script/build-ha-io
+++ b/conf/script/build-ha-io
@@ -230,17 +230,34 @@ net_config_check() {
       die "LNet endpoint $ip2 doesn't appear to be configured at $rnode. $check_msg"
 }
 
+# On both nodes:
+# lvolume is mounted at /var/motr1,
+# rvolume is mounted at /var/motr2.
+mount_var_motr_dirs() {
+    log "${FUNCNAME[0]}: make sure there are no stale mounts hanging..."
+    mountpoint -q /var/motr1 && sudo umount /var/motr1 || true
+    ssh $rnode 'mountpoint -q /var/motr2 && sudo umount /var/motr2 || true'
+
+    log "${FUNCNAME[0]}: check & upgrade old setups with /var/motr directories..."
+    [ -d /var/motr ] && ! [ -d /var/motr1 ] && mv /var/motr /var/motr1
+    ssh $rnode '[ -d /var/motr ] && ! [ -d /var/motr2 ] && mv /var/motr /var/motr2'
+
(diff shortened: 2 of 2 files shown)


conf/script/build-ha-io
```diff
@@ -813,6 +829,7 @@ ha_ops=(
     consul_conf_prepare
     consul_rsc_add
     motr_kernel_rsc_add
+    var_motr_rsc_add
```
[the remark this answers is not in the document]
andriytk: Yeah, let me add a comment about it...

chumakd (approved):

azheregelya (approved):

conf/script/build-cortx-ha
```diff
@@ -27,7 +27,10 @@ init() {
     host2=$(echo $node_list | jq '.["local"][1]' | sed s/\"//g)
 
     # Restart csm to enable decision maker changes
-    pcs resource restart csm-agent
+    cnt=10
```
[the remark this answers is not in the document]
andriytk: The script will stop. But it's a good idea to add some error msg! Will do...

conf/script/build-cortx-ha
```diff
@@ -27,7 +27,10 @@ init() {
     host2=$(echo $node_list | jq '.["local"][1]' | sed s/\"//g)
 
     # Restart csm to enable decision maker changes
```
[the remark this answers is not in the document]
andriytk: Yes, as mentioned in the description - it relates to the issue tried to be fixed by #123.

conf/script/build-ha-io
```diff
@@ -517,6 +527,16 @@ motr_kernel_rsc_add() {
                                            kind=Optional symmetrical=false
 }
 
+var_motr_rsc_add() {
+    log "${FUNCNAME[0]}: Adding var-motr mounts to Pacemaker..."
+    sudo pcs -f $cib_file resource create var-motr1 ocf:heartbeat:Filesystem \
+        device=$lvolume directory=/var/motr1 \
```
mssawant, line 533: @andriytk , now the /var/motr device will be mounted on /var/motr1 and /var/motr2 respectively, do we need to update /etc/sysconfig/m0d-* env files accordingly? as the local motr env file has /var/motr while the guest has /var/motr1 or /var/motr2 as per the node.

andriytk: I thought about it. We can do this - yes. But I decided to leave it with /var/motr symlink for now for the sake of less changes...

mssawant: Okay, also when the node is replaced the directories  won't be present. We need to recreate the directories and symlinks as part of the restore operation.

andriytk: Yes, which function is called on the restore?

mssawant: The functions that need to be invoked for restore need to add "restore" to them as in "bootstrap_update_restore".

andriytk: Updated the patch. Please verify. Thanks!

andriytk: Strange, I updated the branch - (see https://github.com/andriytk/cortx-ha/tree/var-motr), but it is not showing here.

mssawant: not able to see the change yet

andriytk: This is weird. Let me create another PR...

andriytk:
Closed in favour of #144. (Because we could not update the patch here.)

PR #144, opened 2020-09-24
EOS-13393: fix nodes fencing issue on sporadic hax stop timeouts

andriytk:

   after hax is stopped).

Tested manually:
 - build-ha-io PASS;
 - failback/failover PASS;
 - build-ees-ha-update PASS (with a minor fix related to #123).


andriytk:
@mssawant , @chumakd , @azheregelya - this is the new PR instead of #140.

azheregelya (approved):
Looks good.

conf/script/build-ha-io
```diff
@@ -517,6 +524,16 @@ motr_kernel_rsc_add() {
                                            kind=Optional symmetrical=false
 }
 
+var_motr_rsc_add() {
+    log "${FUNCNAME[0]}: Adding var-motr mounts to Pacemaker..."
+    sudo pcs -f $cib_file resource create var-motr1 ocf:heartbeat:Filesystem \
```
mssawant, line 529: need to add this resource to cleanup as well, prov-ha-reset script.

andriytk: I think it will be cleaned as part of c1/2 group there (the same was as hax and others).

mssawant: ah  right!

mssawant (approved):

andriytk:
@ajaysrivas it's ready to be landed.

chumakd (approved):

indrajitzagade (approved):
Approving based on other approvals from hare team.

indrajitzagade:
@andriytk Please fill the PR template going forward.

indrajitzagade:
Please rebase since we cannot merge this yet.

PR #145, opened 2020-09-24
EOS-13356: support different data interface names on the nodes

andriytk:
Currently, the build-ha-io script expects that the name of the
data network interface is the same on both nodes, which sometimes
may not be the case when the underlying HW is different. (Which
may happen, for example, when the old failed node is replaced
with the new one.)

Solution: add support for different interface names via the
two new configuration options: `--left-iface` and `--right-iface`.
If the options are not explicitly set by user, the value of
the old `--iface` option is used for both interfaces.

Tested manually on sm18/19-r20.pun setup.


andriytk:
@mssawant please review it here.

mssawant (approved):

conf/script/build-ha-io
```diff
@@ -65,6 +68,8 @@ or via a yaml file, e.g.:
   ip1: <ip>
   ip2: <ip2>
   interface: <iface>
+  left-iface: <iface>
```
mssawant, line 71: Sorry, it occurred to me that we need to make this change in the provisioner template that generates build-ees-ha-args.yaml and also in prov-m0-reset script used by data recovery team.

andriytk: Where is this template located and what its name? Note: this change is backwards compatible, i.e. it can work with the current Provisioner version without modification.

About prov-m0-reset - right. Will update tomorrow...

mssawant: Currently the args template in provisioner has just 1 interface, need to add another to support different names, both can be same in case the interfaces are same on both the nodes.
cortx-prvsnr/srv/components/ha/iostack-ha/files/ha-params.tmpl

andriytk: I would give the opportunity to Provisioner team for doing this change. cc @ypise

indrajitzagade (approved):

conf/script/build-ha-io
andriytk: @mssawant I don't see why we need to use different interface names in prov-m0-reset - all the commands there are running on the same node... Could you elaborate, please?

rajanikantchirmade: I think `prov-m0-reset` only remove pacemaker resources by resource name and it does not specify any interface names.

I think @mssawant  may be talking related to filesystem resource newly added part of other PR).

rajanikantchirmade: @mssawant  
`prov-m0-reset` does not work on any interface name. 

I think it work when lnet already configured.

indrajitzagade:
@mssawant @andriytk I see still there is some discussion in progress, please let me know when done and the PR is ready for merge.

andriytk:
@indrajitzagade sure.

conf/script/build-ha-io
mssawant: @rajanikantchirmade , @andriytk , we do use iface values to mkfs,

```
mkfs() {
    local node1=$lnode
    local node2=$rnode
    if $single_node; then
        node1=$(/opt/seagate/cortx/hare/libexec/node-name)
        node2=$(/opt/seagate/cortx/hare/libexec/node-name)
    fi

    echo 'Adding ip addresses..'
    run_cmd $node1 "sudo ifconfig $iface:c1 $ip1"
    run_cmd $node2 "sudo ifconfig $iface:c2 $ip2"

    echo 'Adding lnet interfaces..'
    # Its possible that lnet module is not loaded.
    run_cmd $node1 "sudo modprobe lnet"
    run_cmd $node2 "sudo modprobe lnet"

    run_cmd $node1 "sudo lnetctl net add --net $net_type --if $iface:c1"
    run cmd $node2 "sudo lnetctl net add --net $net_type --if $iface:c2"
```

rajanikantchirmade: @mssawant  ohh I got confuse with anther script  `prov-ha-reset`

andriytk: @mssawant but on the same node!

andriytk: @mssawant yes, I see it now... Let me update the patch.

andriytk: @mssawant aren't they run on the same node?

mssawant: sorry, the current script does not require this change. I have a patch in progress which will start mkfs on both the nodes but I will incorporate this change in it later once your patch is landed.

mssawant (approved):

andriytk:
@indrajitzagade it's ready to be landed.

andriytk:
After it is landed, I'll prepare its cherry-pick PR to cortx-1.0.

PR #146, opened 2020-09-25
Problem: prov-m0-reset --mkfs-only does not provide option to run on both nodes

mssawant:
## Problem Statement
<pre>
  <code>
prov-m0-reset --mkfs-only does not provide option to run on both nodes
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
 Tested on hardware setup as part of data recovery.
  </code>
</pre>
## Problem Description
<pre>
  <code>
Due to lvm used for Motr meta data segments running mkfs on the volumes on
single node does not reflect the changes on peer node when data recovery is
executed on both the nodes in parallel. Presently prov-m0-reset --mkfs-only
execution does not provide to run the operation on both the nodes.
  </code>
</pre>
## Solution
<pre>
  <code>
Provide and option for the user to run m0mkfs for both the nodes on single node
or on individual nodes.
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
Run data recovery.
  </code>
</pre>

### DIFF
--- a/conf/script/prov-m0-reset
+++ b/conf/script/prov-m0-reset
@@ -21,6 +21,10 @@ export PS4='+ [${BASH_SOURCE[0]##*/}:${LINENO}${FUNCNAME[0]:+:${FUNCNAME[0]}}] '
 
 PROG=${0##*/}
 
+log_file=${M0_RESET_LOG_FILE:-/var/log/m0-reset.log}
+exec &>> $log_file
+exec &> >(stdbuf -oL gawk '{ print strftime("%Y-%m-%d %H:%M:%S"), $0 }')
+
 usage() {
     cat <<EOF
 Usage: $PROG [OPTS] <CDF> [<params.yaml>]
@@ -35,22 +39,27 @@ Here are some prerequisites/guidelines:
 * The script should be run from the "left" node.
 
 Mandatory parameters:
-  --ip1           <addr>         1st roaming IP address
-  --ip2           <addr>         2nd roaming IP address.
-  --left-node     <n1>  Left  node hostname (default: pod-c1)
-  --right-node    <n2>  Right node hostname (default: pod-c2)
-  --left-volume   <lv>  Left  node /var/motr volume (default: /dev/sdb)
-  --right-volume  <rv>  Right node /var/motr volume (default: /dev/sdc)
-                  <CDF>          Hare Cluster Description File
+  --ip1           <addr> 1st roaming IP address
+  --ip2           <addr> 2nd roaming IP address.
+  --left-iface    <if>  Left node data nw interface (default: eth1)
+  --right-iface   <if>  Right node data nw interface (default: eth1)
+  --left-node     <n1>   Left node hostname (default: pod-c1)
+  --right-node    <n2>   Right node hostname (default: pod-c2)
+  --left-volume   <lv>   Left  node /var/motr volume (default: /dev/sdb)
+  --right-volume  <rv>   Right node /var/motr volume (default: /dev/sdc)
+                  <CDF>  Hare Cluster Description File
 
 Optional parameters:
-  --mkfs-only           run mkfs on motr meta data devices.
+  --mkfs-only            run mkfs on motr meta data devices.
+  --single-node          run mkfs from single node only.
 
 Note: parameters can be specified either directly via cmd-line options
 or via a yaml file, e.g.:
 
   ip1: <ip>
   ip2: <ip2>
+  left-iface: <iface>
+  right-iface: <iface>
   left-node: <lnode>
   right-node: <rnode>
   left-volume: <lvolume>
@@ -59,10 +68,10 @@ EOF
 }
 
 TEMP=$(getopt --options h,i: \
-              --longoptions help,ip1:,ip2: \
+              --longoptions help,ip1:,ip2:,interface:,left-iface:,right-iface: \
               --longoptions left-node:,right-node: \
               --longoptions left-volume:,right-volume: \
-              --longoptions mkfs-only \
+              --longoptions mkfs-only,single-node: \
               --name "$PROG" -- "$@" || true)
 
 (($? == 0)) || { usage >&2; exit 1; }
@@ -72,12 +81,15 @@ eval set -- "$TEMP"
 ip1=
 ip2=
 iface=eth1
+left_iface=
+right_iface=
 net_type=tcp
 lnode=pod-c1
 rnode=pod-c2
 lvolume=/dev/sdb
 rvolume=/dev/sdc
 mkfs_only=
+single_node=false
 
 while true; do
     case "$1" in
@@ -85,12 +97,15 @@ while true; do
         --ip1)               ip1=$2; shift 2 ;;
         --ip2)               ip2=$2; shift 2 ;;
         -i|--interface)      iface=$2; shift 2 ;;
+        --left-iface)        left_iface=$2; shift 2 ;;
+        --right-iface)       right_iface=$2; shift 2 ;;
         --net-type)          net_type=$2; shift 2 ;;
         --left-node)         lnode=$2; shift 2 ;;
         --right-node)        rnode=$2; shift 2 ;;
         --left-volume)       lvolume=$2; shift 2 ;;
         --right-volume)      rvolume=$2; shift 2 ;;
         --mkfs-only)         mkfs_only=true; shift ;;
+        --single-node)       single_node=true; shift 2 ;;
         --)                  shift; break ;;
         *)                   break ;;
     esac
@@ -106,12 +121,19 @@ die() {
     exit 1
 }
 
+log() {
+    local msg=$1
+    echo $msg
+}
+
 # Read Virtual ip, node and meta data volume information from args file.
 if [[ -f $argsfile ]]; then
     while IFS=': ' read name value; do
        case $name in
            ip1)          ip1=$value     ;;
            ip2)          ip2=$value     ;;
+           left-iface)   left_iface=$value   ;;
+           right-iface)  right_iface=$value   ;;
            interface)    iface=$value   ;;
            net-type)     net_type=$value ;;
            left-node)    lnode=$value   ;;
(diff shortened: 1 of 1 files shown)


conf/script/prov-m0-reset
```diff
@@ -244,137 +249,166 @@ get_fid() {
...
-                                  /usr/lib/systemd/system/hare-consul-agent-c2.service
+    run_cmd $node1 "sudo sed -i '/\(#ExecStartPost\|#ExecStopPost\)=\/usr\/sbin\/attrd_updater/s/#//g' \
+                        /usr/lib/systemd/system/hare-consul-agent-c1.service"
+    run_cmd $node2 "sudo sed -i '/\(#ExecStartPost\|#ExecStopPost\)=\/usr\/sbin\/attrd_updater/s/#//g' \
+                        /usr/lib/systemd/system/hare-consul-agent-c2.service"
 
-    sudo systemctl daemon-reload
-    sudo systemctl reset-failed
+    run_cmd $node1 "sudo systemctl daemon-reload"
+    run_cmd $node2 "sudo systemctl reset-failed"
```
indrajitzagade, line 407: these commands need to be run on both servers @mssawant ?

mssawant: @indrajitzagade , now, by default the commands are executed on both the nodes but there's a single node option as well to run the commands from singlenode for both the nodes.

indrajitzagade:
> ## Problem Statement
>   ```
> 
> prov-m0-reset --mkfs-only does not provide option to run on both nodes
>   
> ```
> 
> 
> ## Unit testing on RPM done
>   ```
> 
>   No
>   
> ```
> 
> 
> ## Problem Description
>   ```
> 
> Due to lvm used for Motr meta data segments running mkfs on the volumes on
> single node does not reflect the changes on peer node when data recovery is
> executed on both the nodes in parallel. Presently prov-m0-reset --mkfs-only
> execution does not provide to run the operation on both the nodes.
>   
> ```
> 
> 
> ## Solution
>   ```
> 
> Provide and option for the user to run m0mkfs for both the nodes on single node
> or on individual nodes.
>   
> ```
> 
> 
> ## Unit Test Cases
>   ```
> 
> TBD
>   
> ```

I believe @shashank-parulekar is testing these changes. @shashank-parulekar Please let us know here after your testing and complete the review accordingly.

indrajitzagade:
Waiting for the test results and review from hare team.

conf/script/prov-m0-reset
madhavemuri: systemd reload and reset-failed are not being run on both the nodes.
They need to be run on both the nodes.
​

conf/script/prov-m0-reset
```diff
@@ -130,10 +152,32 @@ fi
 [[ -b $lvolume ]] || die "meta-data volume $lvolume is not available"
 [[ -b $rvolume ]] || die "meta-data volume $rvolume is not available"
 
+export PATH=$PATH:/opt/seagate/cortx/hare/bin
+
+# This function is duplicated from build-ha-io script.
+# TODO: Remove the function definition once the common functions are available.
+prepare_var_motr_dirs() {
+    log "${FUNCNAME[0]}: make sure there are no stale mounts hanging..."
+    mountpoint -q /var/motr1 && sudo umount /var/motr1 || true
```
madhavemuri, line 161: What if this script is run from right node ?

conf/script/prov-m0-reset
```diff
@@ -244,137 +284,214 @@ get_fid() {
...
-    systemctl is-active --quiet m0d@$ios_fid_c2 &&
+    run_cmd $node1 "systemctl is-active --quiet m0d@$confd_fid_c1" ||
+    run_cmd $node1 "systemctl is-active --quiet m0d@$ios_fid_c1" ||
+    run_cmd $node2 "systemctl is-active --quiet m0d@$confd_fid_c2" ||
+    run_cmd $node2 "systemctl is-active --quiet m0d@$ios_fid_c2" &&
         die 'Active m0d instances found, please stop all the m0d instances'
 
-    sudo systemctl start motr-mkfs@$confd_fid_c1
+    run_cmd $node1 "sudo systemctl start motr-mkfs@$confd_fid_c1"
     while [[ $(get_proc_state $confd_fid_c1) != M0_CONF_HA_PROCESS_STOPPED ]]; do
```
madhavemuri, line 351: What is there is a consul failure and consul is not accessible, how it is being handled ?

andriytk:
> Due to lvm used for Motr meta data segments running mkfs on the volumes on
single node does not reflect the changes on peer node when data recovery is
executed on both the nodes in parallel.

Sorry - why/when is it needed to execute it on both nodes in parallel?

conf/script/prov-m0-reset
mssawant: Yes, for reset it was assumed that is the node is not working then that is the first problem that we need to work on. But its still a valid case but again I will make this change for reset separately as there is other related code for reset that needs to be updated for multinode case.
I have updated mkfs for now which is being used. Do you want to fix reset in this patch too?

conf/script/prov-m0-reset
mssawant: We make sure that consul has started successfully in mkfs() function. We are waiting for its leader to be elected and only then proceed with starting further services. If Consul fails here then its need to be addressed separately.

conf/script/prov-m0-reset
mssawant: Good catch, let me fix this.

conf/script/prov-m0-reset
mssawant: Also another reason for reset to be executed from primary node as it runs hctl bootstrap which requires some files which are presently only created on primary node.

andriytk:
> So if the mkfs is done for both the lvms from single node then due to lvm limitations the mkfs done for the remote lvm is not reflected.

This is interesting... Can you elaborate a bit more on this? What kind of limitations?

cc @nikitadanilov, @max-seagate.

mssawant:
@andriytk Okay, so its not about motr-mkfs but lvm itself. We are using lvms which are accessible by both the nodes. But if you are writing to remote lvm looks like the lvm needs to be shutdown on the peer node, (https://tldp.org/HOWTO/LVM-HOWTO/sharinglvm1.html). Else somehow the changes weren't reflected on the remote lvm. This is what was observed by @shashank-parulekar . So the script was mounting both the lvms on single node and running mkfs over it but the mkfs  done to the remote lvm were not seen when the lvm was also active on the remote node. I have not tested this personally, this was reported by the data recovery team. Now that we are running mkfs from both the nodes, mkfs results are reflected on both the lvms as per data recovery team.
I suppose for this we need https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/logical_volume_manager_administration/lvm_cluster_overview may be. @max-seagate can add more details to it probably.

shashank-parulekar:
> > So if the mkfs is done for both the lvms from single node then due to lvm limitations the mkfs done for the remote lvm is not reflected.
> 
> This is interesting... Can you elaborate a bit more on this? What kind of limitations?
> 
> cc @nikitadanilov, @max-seagate.

@andriytk The recovery tool while launched from one node tries to run recovery operations on both the nodes in parallel if both the nodes are UP. If the remote node is not running or the remote storage cannot be accessed from the remote node then the recovery falls back to completing the work from the local node.
For Recovery we create a Read-only LVM snapshot of the MD volume (on both the nodes)so that the Recovery instance running on each node can use this Snapshot to read the original data; the 'recovered data' is written to the original MD. The Read-only LVM Snapshot feature is implemented using COW. Now when we have created the Read-only LVM snapshot for the remote node as long as the Writes to the original MD volume are done from the same node everything works fine because the LVM module running on the remote node implements the COW feature as it knows which locations of the MD volume are re-written. Unfortunately when we do mkfs for the remote node's storage from local node the mkfs data which is written to the remote node's MD volume is not tracked by the LVM module running on the remote node as it has no way to know these changes. Since the remote node's LVM module is not able to do the COW feature, as this data was written by local node, the orignal copy for Read-only access is not COW'ed and now the newly minted mkfs Volume data (from the local node) is seen through the Snapshot interface on the Remote node. 
In short this affects only LVM snapshot feature because of the above mentioned reason.

shashank-parulekar (approved):

madhavemuri (approved):

indrajitzagade (approved):
Approving and merging based on approvals from @shashank-parulekar  and @madhavemuri

saurabhkhanvilkar (approved):

andriytk:
Ok, so will it work on both nodes and on one node (when the other is not available) from now?

conf/script/prov-m0-reset
```diff
@@ -59,10 +68,10 @@ EOF
 }
 
 TEMP=$(getopt --options h,i: \
-              --longoptions help,ip1:,ip2: \
+              --longoptions help,ip1:,ip2:,interface:,left-iface:,right-iface: \
```
azheregelya, line 71: `interface` parameter is not documented, but processed.

conf/script/prov-m0-reset
```diff
@@ -35,22 +39,27 @@ Here are some prerequisites/guidelines:
...
-  --right-node    <n2>  Right node hostname (default: pod-c2)
-  --left-volume   <lv>  Left  node /var/motr volume (default: /dev/sdb)
-  --right-volume  <rv>  Right node /var/motr volume (default: /dev/sdc)
-                  <CDF>          Hare Cluster Description File
+  --ip1           <addr> 1st roaming IP address
+  --ip2           <addr> 2nd roaming IP address.
+  --left-iface    <if>  Left node data nw interface (default: eth1)
+  --right-iface   <if>  Right node data nw interface (default: eth1)
+  --left-node     <n1>   Left node hostname (default: pod-c1)
+  --right-node    <n2>   Right node hostname (default: pod-c2)
```
andriytk, line 47: Minor indentation issue (space between <if> and the next text).

conf/script/prov-m0-reset
andriytk: Not critical anymore as we are going to use two parameters for possible different interface names cases. We just can keep the old parameter for some time just for the sake of compatibility with Provisioner and just drop it later completely.

PR #150, opened 2020-09-25
EOS-13393: fix nodes fencing issue on sporadic hax stop timeouts (#144)

andriytk:

   after hax is stopped).

Tested manually:
 - build-ha-io PASS;
 - failback/failover PASS;
 - build-ees-ha-update PASS (with a minor fix related to #123).

Reviewed-by: Mandar Sawant <mandar.sawant@seagate.com>
Signed-off-by: Andriy Tkachuk <andriy.tkachuk@seagate.com>
(cherry picked from commit 7b7197cd8f2ec3a53dd7166ec4e4d2d6741d6e30)

### DIFF
--- a/conf/script/build-ha-io
+++ b/conf/script/build-ha-io
@@ -230,17 +230,32 @@ net_config_check() {
       die "LNet endpoint $ip2 doesn't appear to be configured at $rnode. $check_msg"
 }
 
+# On both nodes:
+# lvolume is mounted at /var/motr1,
+# rvolume is mounted at /var/motr2.
+prepare_var_motr_dirs() {
+    log "${FUNCNAME[0]}: make sure there are no stale mounts hanging..."
+    mountpoint -q /var/motr1 && sudo umount /var/motr1 || true
+    ssh $rnode 'mountpoint -q /var/motr2 && sudo umount /var/motr2 || true'
+
+    log "${FUNCNAME[0]}: check & upgrade old setups with /var/motr directories..."
+    [ -d /var/motr ] && ! [ -d /var/motr1 ] && mv /var/motr /var/motr1
+    ssh $rnode '[ -d /var/motr ] && ! [ -d /var/motr2 ] && mv /var/motr /var/motr2'
+
+    log "${FUNCNAME[0]}: prepare dirs..."
+    run_on_both 'mkdir -p /var/motr{1,2}'
+    ln -sf /var/motr1 /var/motr
+    ssh $rnode 'ln -sf /var/motr2 /var/motr'
+
+    sudo mount $lvolume /var/motr1 || true
+    ssh $rnode "sudo mount $rvolume /var/motr2 || true"
+}
+
 bootstrap() {
     if ! $skip_mkfs; then
         sudo mkfs.ext4 -q $lvolume >/dev/null <<< y
         sudo mkfs.ext4 -q $rvolume >/dev/null <<< y
     fi
-
-    # Mount /var/motr (if not mounted).
-    mkdir -p /var/motr &&
-      ! mountpoint -q /var/motr && sudo mount $lvolume /var/motr || true
-    ssh $rnode "mkdir -p /var/motr &&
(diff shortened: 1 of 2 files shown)


indrajitzagade (approved):

indrajitzagade:
@mssawant @ajay-paratmandali Please review.
@andriytk We will merge this after there are sufficient approvals

indrajitzagade:
> Currently, /var/motr is mounted before hax service startup and
> unmounted after hax service stop. This mounting/unmounting is done
> in hax systemd service unit file. So when hax fails to stop within
> 30 secs timeout (configured in Pacemaker), the node will be fenced
> by Pacemaker to avoid the situation when /var/motr is mounted on
> both nodes at the same time (which would lead to data corruption).
> 
> Hax stopping can be delayed for different reasons, mainly due to
> bugs in Motr which are not easy to fix quickly (for example, hax
> calls rconfc_start in fs-stats thread and it may get stuck if confd
> process dies and does not come up during the node shutdown). So
> we are solving it from another side:
> 
> 1. Create separate resource in Pacemaker for /var/motr mgmt and
>    stop doing it from hax systemd unit file.
> 2. Add this resource into the Motr Pacemaker group just before hax
>    so that /var/motr mount is ready before hax startup (and unmounts
>    after hax is stopped).
> 3. Set 5 seconds systemd stop timeout for hax (in similar way to
>    all the rest Motr systemd services stop timeout) so that hax is
>    killed by systemd if it cannot stop within this timeframe (for
>    whatever reason) and it's not affecting the cluster state.
> 
> Tested manually:
> 
> * build-ha-io PASS;
> * failback/failover PASS;
> * build-ees-ha-update PASS (with a minor fix related to #123).
> 
> Reviewed-by: Mandar Sawant [mandar.sawant@seagate.com](mailto:mandar.sawant@seagate.com)
> Signed-off-by: Andriy Tkachuk [andriy.tkachuk@seagate.com](mailto:andriy.tkachuk@seagate.com)
> (cherry picked from commit [7b7197c](https://github.com/Seagate/cortx-ha/commit/7b7197cd8f2ec3a53dd7166ec4e4d2d6741d6e30))

@andriytk  Can you please fill the PR template?

andriytk:
The template is ugly and inconvenient. It's a waste of the time to fill it and to use it. Please change the template so it at least fits the screen and that all the parts in it can be understood by someone.

andriytk:
@indrajitzagade , as mentioned in the description above - this patch is just

> (cherry picked from commit 7b7197c)

So there is no need for everyone to review it again. I think. (Especially there are at least 5 blocker tickets waiting for it.)

indrajitzagade:
@andriytk We need at least two approvals.
@saurabhkhanvilkar @ajay-paratmandali  Can you please review?

indrajitzagade:
@andriytk Regarding the PR template, I will discuss with @ajaysrivas

conf/script/build-ha-io
```diff
@@ -518,6 +525,16 @@ motr_kernel_rsc_add() {
                                            kind=Optional symmetrical=false
 }
 
+var_motr_rsc_add() {
+    log "${FUNCNAME[0]}: Adding var-motr mounts to Pacemaker..."
```
rajanikantchirmade, line 529: Sorry, I am late for review.

I think we need to delete resource from `prov-ha-reset`

Otherwise sw update will failed

andriytk: I think this was discussed here already - https://github.com/Seagate/cortx-ha/pull/144#discussion_r494450608. But it was a good spot! ;)

mssawant: In prov-ha-reset we delete the entire c1,c2 group so the resource must get deleted as part of that.

mssawant (approved):

conf/script/build-ha-io
rajanikantchirmade: ohh got it. 
Thanks

rajanikantchirmade (approved):

PR #164, opened 2020-09-26
build-ha-io: fix bogus failure of prepare_var_motr_dirs()

andriytk:
Currently, the commands which check for /var/motr dir present
at prepare_var_motr_dirs() (and run some upgrade logic if it
is) do not tolerate the situation when /var/motr directory
is not present - they would return non-zero status (thus,
failing the whole script which is wrong).

Solution: fix the commands by appending `|| true` to them.

Kudos to Mandar Sawant who caught this.


andriytk:
This is the same fix as #163, but to the main branch.

@mssawant please review.

mssawant (approved):

saurabhkhanvilkar:
Same as #163 but for main branch. #163 is already approved and merged. Using my admin privileges to merge this one into main branch.

saurabhkhanvilkar (approved):

indrajitzagade (approved):

PR #167, opened 2020-09-28
Problem: wrong /var/motr links created if restore executed from secondary node

mssawant:
## Problem Statement
<pre>
  <code>
 Problem: wrong /var/motr links created if restore executed from secondary node
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  No
  </code>
</pre>
## Problem Description
<pre>
  <code>
prepare_var_motr_dirs() is also executed to restore /var/motr directories
post node replacement. But the function presently assumes it to be executed
from primary node only which may not be the case during restore as restore
can be executed from secondary node as well. This creates wrong /var/motr
links.
  </code>
</pre>
## Solution
<pre>
  <code>
Run /var/motr creation commands over ssh for both the nodes to handle the
execution from any of the nodes.
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
 None
  </code>
</pre>

PR #168, opened 2020-09-29
Problem: wrong /var/motr links created if restore executed from secon…

mssawant:
## Problem Statement
<pre>
  <code>
Problem: wrong /var/motr links created if restore executed from secondary node
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  Yes
  </code>
</pre>
## Problem Description
<pre>
  <code>
prepare_var_motr_dirs() is also executed to restore /var/motr directories
post node replacement. But the function presently assumes it to be executed
from primary node only which may not be the case during restore as restore
can be executed from secondary node as well. This creates wrong /var/motr
links.
Also motr systemd unit file is not updated on restore.
  </code>
</pre>
## Solution
<pre>
  <code>
- Run /var/motr creation commands over ssh for both the nodes to handle the
  execution from any of the nodes.
- Update motr systemd unit on restore as well.
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
 Server failback must work post node restore operation.
  </code>
</pre>

### DIFF
--- a/conf/script/build-ha-io
+++ b/conf/script/build-ha-io
@@ -260,19 +260,19 @@ net_config_check() {
 # rvolume is mounted at /var/motr2.
 prepare_var_motr_dirs() {
     log "${FUNCNAME[0]}: make sure there are no stale mounts hanging..."
-    mountpoint -q /var/motr1 && sudo umount /var/motr1 || true
+    ssh $lnode 'mountpoint -q /var/motr1 && sudo umount /var/motr1 || true'
     ssh $rnode 'mountpoint -q /var/motr2 && sudo umount /var/motr2 || true'
 
     log "${FUNCNAME[0]}: check & upgrade old setups with /var/motr directories..."
-    [ -d /var/motr ] && ! [ -d /var/motr1 ] && mv /var/motr /var/motr1 || true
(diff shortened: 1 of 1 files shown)


andriytk (approved):

indrajitzagade (approved):
I think this is merge to main branch from cortx-1.0 branch.

andriytk:
It is, yes.

conf/script/build-ha-io
```diff
@@ -260,19 +260,19 @@ net_config_check() {
 # rvolume is mounted at /var/motr2.
 prepare_var_motr_dirs() {
     log "${FUNCNAME[0]}: make sure there are no stale mounts hanging..."
-    mountpoint -q /var/motr1 && sudo umount /var/motr1 || true
+    ssh $lnode 'mountpoint -q /var/motr1 && sudo umount /var/motr1 || true'
```
ajaysrivas, line 263: why not use run_on_both function for all of such commands?

andriytk: Because the commands are different?

andriytk:
This is the merge of #167 to the main branch. (Just for the record.)

PR #172, opened 2020-09-29
Problem: cob domain is unavailable even after motr ios is started

mssawant:
## Problem Statement
<pre>
  <code>
cob domain is unavailable even after motr ios is started
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  Yes, on hardware system.
  </code>
</pre>
## Problem Description
<pre>
  <code>
After motr-mkfs, motr ioservices are needed to be started to facilitate cob
domain creation which is not created as part of mkfs operation.
But still cob domain is not available even after motr ioservice reported
M0_CONF_HA_PROCESS_STARTED.
  </code>
</pre>
## Solution
<pre>
  <code>
- Wait for sometime (15 secs) before stopping motr ioservices to allow
  some time for cob domain creation.
- Fix typo and invoke cleanup before checking for consul service at mkfs
  start.
  Note: Created https://jts.seagate.com/browse/EOS-13779 to remove this
            workaround once https://jts.seagate.com/browse/EOS-13778 is fixed.
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
Tested on hardware, script completed successfully.
  </code>
</pre>

### DIFF
--- a/conf/script/prov-m0-reset
+++ b/conf/script/prov-m0-reset
@@ -336,6 +336,27 @@ cleanup() {
     run_cmd $node2 "mountpoint -q /var/motr2 && sudo umount /var/motr2 || true"
 }
 
+has_m0d_started() {
+    fid=$1
+    node=$2
+    
+    while [[ $(run_cmd $node "systemctl is-active m0d@$fid") != 'active' ]]; do
+       sleep 5
+    done
+
+    while [[ $(get_proc_state $fid) != M0_CONF_HA_PROCESS_STARTED ]]; do
+        sleep 1
+    done
+}
+
+has_m0d_stopped() {
+    fid=$1
+
+    while [[ $(get_proc_state $fid) != M0_CONF_HA_PROCESS_STOPPED ]]; do
+        sleep 1
+    done
+}
+
 m0_mkfs() {
     local confd_fid_c1=$(get_fid 'confd' $lnode)
     local ios_fid_c1=$(get_fid 'ios' $lnode)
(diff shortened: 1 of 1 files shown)


shashank-parulekar (approved):

madhavemuri (approved):

andriytk:
Why do we need to create the cob domains also? Why can't they be just created later when m0d processes will be started?

mssawant:
@andriytk yes, that is what is missing in motr. Have asked @madhavemuri and @shashank-parulekar to create respective tickets.
cc @max-seagate

andriytk:
@mssawant sorry, I'm not sure I understood - what is missing in Motr? You create cobs by starting m0d processes, right? So why this can't be done later when m0d will be started? Why is this needed to be done in this script explicitly my starting m0d processes after m0mkfs?

andriytk:
> Here for data recovery, which is done offline the cob domains are needed.

Do you mean that there is some another program (not m0d) which does data recovery and this program needs the cobs and it does not create them by itself?

mssawant:
@andriytk data recovery is done offline and it requires mkfs to be done on the /var/motr volumes. Now after mkfs the recovery process expects all the meta data related objects to be present including cob domain, not cobs. COB domain is presently not created during mkfs, which ideally should be. COB domain is created when ioservice is started. So that is what we are doing here. But this needs to be fixed in motr that COB domain needs to be created during mkfs itself.
Another issue here is that COB domain is somehow not available even after ioservice has reported M0_HA_PROCESS_STARTED to Hare and needs more time, hence the sleep for now. This also needs to be investigated and fixed.
cc @shashank-parulekar

andriytk:
Do you say it's a quick temporary hack which should be deleted later? If so - should we create a Jira ticket mentioning this so that it's not forgotten?

mssawant:
@andriytk Yes, this is as suggested by data recovery team. I have added the comment as well in the script and have asked @shashank-parulekar and @madhavemuri to create the respective tickets.

andriytk:
Ok. Please add `XXX` to your comments mentioning that this is a temporary code which should be removed after this and this Jira tickets (specify them there) are fixed. And please create another Jira ticket for this (removal) activity and link it with the Motr tickets as "Depend On".

mssawant:
@andriytk do you think it would be good to mention this in the commit message?

andriytk:
Yes. And I hope this code will be removed soon.

andriytk:
Yes - add this to the commit-msg also, please. (Not instead, but in addition.)

andriytk (approved):
I approve this patch only under mentioned above stipulations.

andriytk:
And if it is possible to *not* merge it to the main branch - I'd prefer to not do it.

mssawant:
@andriytk created https://jts.seagate.com/browse/EOS-13779 to track workaround and https://jts.seagate.com/browse/EOS-13778 from motr side to fix this. Updated commit message.

shashank-parulekar:
> And if it is possible to _not_ merge it to the main branch - I'd prefer to not do it.

@andriytk - Why do we not want to merge to 'main' ? It might be a hack for now and we need to work on fixing it but I am not sure we have plans to do that by GA. Without this change in main - you have Data Recovery working on cortx-1.0 but failing (actually crashing) on code in 'main' !!!.
Since the tickets have been created let us add it now to 'main' branch and the fixes (whenever done) will be put in both the branches.
I agree it might not look pretty but the behavior is similar to running 'hctl bootstrap --mkfs' followed by 'hctl shutdown'.

saurabhkhanvilkar (approved):

indrajitzagade (approved):
approving and merging based on approvals from others

PR #180, opened 2020-10-02
EOS-13835: fix exec permission for prov-m0-reset

andriytk:
Executable permission should be set for the script. Otherwise, data recovery fails.


azheregelya (approved):

knekrasov (approved):

mssawant (approved):

ajaysrivas:
Please use pull request template while creating a PR.

ajaysrivas (approved):

andriytk:
> Please use pull request template while creating a PR.

@ajaysrivas , the template is bad and awkward - that's why I prefer to not use it. Please update the template so has only the meaningful clauses (or at least self-evident) and make it to look good (it should not take a few pages only by itself when there is no text it in).

Thank you.

ajaysrivas:
@andriytk Modifications are welcome to the template. I am not particular about the format but everyone should follow same format and it should contain following information -
1. Description of problem
2. Explanation of solution
3. Unit/Integration tests executed.
4. Any other information which we may think relevant as code reviewer/gatekeeper.

andriytk:
@ajaysrivas I agree. Looks like we pursuing the same purpose here.

What I'm saying is that it looks like the current template does not serve the purpose - people are ignoring filling it and gatekeepers are merging patches with it ignoring the fact that there is no useful information present there.

As for me - it is better not having any template at all. Because it will be very obvious that something is missing and it will urge gatekeepers to ask questions about the patch and ask people to describe it rather than having empty template in the description which makes a false impressing that there might be something useful in it.

ajaysrivas:
@andriytk If the information is available before hand then it will save time both for gatekeepers as well as PR submitter. Asking same questions on every PR will increase turn around time and I want to save that time by enforcing a template which answers required question for every PR. It's only matter of cultivating the habit. Reviewers need to remind every PR submitter for some time.

andriytk:
One more thing: when PR is created - github automatically takes the commit-msg body into the description. This is very handy and saves time a lot. That's another reason why templates are useless in PRs. (They might be handy when creating issues though.)

All the useful information that might be in the PR description should be also available in the git commit-msg. So it's better to cultivate the habit of writing good commit messages - that's where it all begins...

PR #185, opened 2020-10-05
sspl-ra: fix monitor result in master mode

andriytk:
In master mode, when sspl resource is not running, the resulting
code of monitor command returned by sspl resource agent would be
OCF_ERR_GENERIC which is wrong. Upon receiving such a code
Pacemaker would not try to restart the resource.

Solution: return OCF_NOT_RUNNING code instead.

To facilitate debugging, we add logging into the syslog also,
as a bonus.

Tested manually on the setup while debugging the problem
about SSPL promotion failure.

### DIFF
--- a/ha/resource/sspl
+++ b/ha/resource/sspl
@@ -80,6 +80,10 @@ END
     exit $OCF_SUCCESS
 }
 
+log() {
+    logger -t sspl-ra "$*"
+}
+
 stateful_usage() {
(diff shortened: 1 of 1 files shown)


indrajitzagade (approved):

Madhura-08 (approved):
Looks good

saurabhkhanvilkar (approved):

PR #191, opened 2020-10-05
EOS-13296: haproxy service does not restart automatically on failure

mssawant:
Solution:

## Problem Statement
<pre>
  <code>
haproxy service does not restart automatically on failure
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  Not rpm but tried the command on LDR1 hw setup.
  </code>
</pre>
## Problem Description
<pre>
  <code>
If a pacemaker resource fails, pacemaker will try to restart the resource but if the
resource fails repeatedly then the failcount set by the pacemaker for the resource
needs to be cleared using `pcs resource cleanup`.
Setting failure-timeout meta attribute allows failcounts to expire automatically so
pacemaker tries to restart the resource.
  </code>
</pre>
## Solution
<pre>
  <code>
Set failure-timeout meta attribute of the haproxy pacemaker resource inorder to allow
automatic expiration of haproxy failcounts to trigger restart.
Note: Resource start and stop failures are treated as fatal by default by pacemaker and
in which case the resource failcount is set to -INFINITY and cannot be reset by
failure-timeout.
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
- fail haproxy pacemaker resource
- verify it is failed in pacemaker
- pacemaker will try to restart haproxy after 10s.
  </code>
</pre>

### DIFF
--- a/conf/script/build-ha-io
+++ b/conf/script/build-ha-io
@@ -738,8 +738,10 @@ statsd_rsc_add() {
 
 haproxy_rsc_add() {
     log "${FUNCNAME[0]}: Adding haproxy to pacemaker..."
-    sudo pcs -f $cib_file resource create haproxy-c1 systemd:haproxy op monitor interval=30s
-    sudo pcs -f $cib_file resource create haproxy-c2 systemd:haproxy op monitor interval=30s
+    sudo pcs -f $cib_file resource create haproxy-c1 systemd:haproxy op monitor interval=30s \
+        meta failure-timeout=10s
+    sudo pcs -f $cib_file resource create haproxy-c2 systemd:haproxy op monitor interval=30s \
+        meta failure-timeout=10s
     sudo pcs -f $cib_file constraint location haproxy-c1 prefers $lnode=INFINITY
     sudo pcs -f $cib_file constraint location haproxy-c1 avoids $rnode=INFINITY
     sudo pcs -f $cib_file constraint location haproxy-c2 prefers $rnode=INFINITY


conf/script/build-ha-io
```diff
@@ -738,8 +738,10 @@ statsd_rsc_add() {
 
 haproxy_rsc_add() {
     log "${FUNCNAME[0]}: Adding haproxy to pacemaker..."
-    sudo pcs -f $cib_file resource create haproxy-c1 systemd:haproxy op monitor interval=30s
-    sudo pcs -f $cib_file resource create haproxy-c2 systemd:haproxy op monitor interval=30s
+    sudo pcs -f $cib_file resource create haproxy-c1 systemd:haproxy op monitor interval=30s \
```
ajaysrivas, line 741: Repeated failures means something is wrong with haproxy and needs manual intervention. Does clearing failure count automatically means keep on trying infinitely? If yes, then why do we need it and what is the case it's handling?

ajaysrivas:
Please pass the DCO.

conf/script/build-ha-io
mssawant: >Does clearing failure count automatically means keep on trying infinitely?
haproxy resources are active/active and anyway are confined to a node, pacemaker anyway tries to restart a failed resource, failure-timeout will just help to clear the failcount and avoid the resource to be banned if failcount reaches 100000 (which high, but can occur during testing, as in case of EOS-13296).

So, if there's some problem with haproxy (e.g. configuration or dependency) itself then we must look into it. But say if haproxy started but failed for some reason (some transient network issue, minor configuration problem post startup), pacemaker does try to restart it anyway but the failcount is not cleared, setting failure-timeout will clear the failcount as well, typically the failcount has to reach 100000 for the resource to be banned.
Note that start/stop failures are still fatal, i.e. pacemaker will set the failcount to -INFINITY if a resource fails to start/stop, this is not cleared by failure-timeout automatically and requires some additional global properties to be set, start-failure-is-fatal (default true) or cluster-recheck-interval (default 15 mins). I think we should not touch the global properties yet.

rajanikantchirmade (approved):

mssawant:
As per the discussions and findings,
https://jts.seagate.com/browse/EOS-13296?focusedCommentId=1953808&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1953808
and
https://jts.seagate.com/browse/EOS-13296?focusedCommentId=1953901&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1953901
looks like the fix is not needed. This is a good to have fix but let's not fix if its not broken for now.
Closing the PR.

PR #195, opened 2020-10-07
EOS-13330: huge stop-timeout for s3srv adds delay for stopping

knekrasov:
## Problem Statement
<pre>
  <code>
  Story Ref (if any): https://jts.seagate.com/browse/EOS-13330
We need to improve failover/failback speed.
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  No
  </code>
</pre>
## Problem Description
<pre>
  <code>
    Huge stop-timeout for S3 server can add unnecessary delays during failover. The reason is that when Pacemaker stops 
systemd-based resource, it needs to start polling its status to ensure that the resource is stopped for sure. According to 
Pacemaker sources, the delay between those polling iterations depends on timeout value: delay = min(2000, timeout / 10). So if 
timeout is too big, resource status will also be checked rarely.
  </code>
</pre>
## Solution
<pre>
  <code>
According to S3 team, it is quite safe to assume 15-second timeout for stop operations.
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>N/A</code>
</pre>


ajaysrivas (approved):

andriytk:
> delay = min(2000, timeout / 10). So if 
timeout is too big, resource status will also be checked rarely.

@knekrasov isn't it 2 seconds interval for checking?

knekrasov:
Yes, it seems like at maximum the delay granularity will be 2 seconds.

```c
#ifdef PCMK__TIME_USE_CGT
    if (goagain) {
        int time_sum = time_diff_ms(NULL, &(cmd->t_first_run));
        int timeout_left = cmd->timeout_orig - time_sum;
        int delay = cmd->timeout_orig / 10;

        if(delay >= timeout_left && timeout_left > 20) {
            delay = timeout_left/2;
        }

        delay = QB_MIN(2000, delay);
        if (delay < timeout_left) {
            cmd->start_delay = delay;
            cmd->timeout = timeout_left;

            if(cmd->exec_rc == PCMK_OCF_OK) {
                crm_debug("%s %s may still be in progress: re-scheduling (elapsed=%dms, remaining=%dms, start_delay=%dms)",
                          cmd->rsc_id, cmd->real_action, time_sum, timeout_left, delay);

            } else if(cmd->exec_rc == PCMK_OCF_PENDING) {
                crm_info("%s %s is still in progress: re-scheduling (elapsed=%dms, remaining=%dms, start_delay=%dms)",
                         cmd->rsc_id, cmd->action, time_sum, timeout_left, delay);

            } else {
                crm_notice("%s %s failed '%s' (%d): re-scheduling (elapsed=%dms, remaining=%dms, start_delay=%dms)",
                           cmd->rsc_id, cmd->action, services_ocf_exitcode_str(cmd->exec_rc), cmd->exec_rc, time_sum, timeout_left, delay);
            }

            cmd_reset(cmd);
            if(rsc) {
                rsc->active = NULL;
            }
            schedule_lrmd_cmd(rsc, cmd);

            /* Don't finalize cmd, we're not done with it yet */
            return;

        } else {
            crm_notice("Giving up on %s %s (rc=%d): timeout (elapsed=%dms, remaining=%dms)",
                       cmd->rsc_id, cmd->real_action?cmd->real_action:cmd->action, cmd->exec_rc, time_sum, timeout_left);
            cmd->lrmd_op_status = PCMK_LRM_OP_TIMEOUT;
            cmd->exec_rc = PCMK_OCF_TIMEOUT;
            cmd_original_times(cmd);
        }
    }
#endif

```

indrajitzagade (approved):

chumakd (approved):

conf/script/build-ha-io
```diff
@@ -817,8 +817,14 @@ _s3server_rsc_add() {
...
    for i in ${s3server_fids[@]}; do
+      # [KN] According to S3 team, 15 seconds is sufficient for shutting down
+      # Moreover, in s3server systemd the timeout is 7sec so 15sec at Pacemaker should suffice.
+      #
+      # As a side effect, this will force Pacemaker to poll status (while stopping) more often.
+      # According to Pacemaker sources, systemd resources started/stopped and then monitored during 'timeout' 
+      # period until the desired status is reported. Delay between those 'monitor' actions is min(2000, timeout / 10).
       sudo pcs -f $cib_file resource create s3server-$suffix-$count systemd:$i op \
-          stop timeout=600
+          stop timeout=15
```
andriytk, line 827: If (for some reason) s3server won't be able to stop in 15 secs - Pacemaker will fence the node. Is this what we want?

knekrasov: At systemd level [there is another timeout of 7 seconds](https://github.com/Seagate/cortx-ha/blob/4e59c673f8db0c06c6d5a54e2d6bcb31c7891241/conf/script/build-ha-io#L775) for s3server. Do you think there is a chance that systemd would not manage to kill the process within 15 seconds?

andriytk: Ok then. But still - what will be the polling timeout now after we use 15 seconds? 15/10 == 1.5 seconds instead of 2? What's the percentage that we gain with this for the total failover/back time?

knekrasov: Perhaps we need to experiment with this setting first.

ajaysrivas:
This should be merged after testing is done. @knekrasov let us know when testing is complete.

azheregelya:
According to discussions the change does not make much sense. It also affects R1 only so far.

PR #203, opened 2020-10-13
EOS-13932: HA: Build 146 : After single controller shutdown, failover…

Madhura-08:
… happening to secondary node

- Main branch PR
- Remove check to detect whether system is cross-connect or not
- Load cross-connect rule file by default

## Problem Statement
<pre>
  <code>
  Story Ref (if any):
    EOS-13932
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  Yes/No
  </code>
</pre>
## Problem Description
<pre>
  <code>
        Stack failover happened after controller fault which was unexpected. The problem is that cortx-HA ends-up installing non- 
        cross connect rules engine schema file on a cross-connect setup since the provisioner has removed a logic to detect and 
        convey this information to other modules e.g. cortx-HA. So till this gets fixed in provisioner, cortx-HA will assume that it is 
        a cross-connect setup.

  </code>
</pre>
## Solution
<pre>
  <code>
   Remove check to detect whether system is cross-connect or not
   Load cross-connect rule file by default
   It is a straight-forward fix to always use cross-connect rule file  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
    Unit Testing details here...
  </code>
</pre>

### DIFF
--- a/conf/script/ha_setup
+++ b/conf/script/ha_setup
@@ -49,7 +49,7 @@ config() {
     mkdir -p ${HA_CONF}
 
     # TODO Add generation for rule engine to make rule engine schama
-    rule_engine_config
+    cp -rf ${SOURCE_HA_CONF}/rules_engine_schema_jbod.json ${HA_CONF}/${HA_CONF_RULE_FILE}
 
     # ha conf
     cp -rf ${SOURCE_HA_CONF}/ha.conf ${HA_CONF}/
--- a/conf/script/ha_setup
+++ b/conf/script/ha_setup
@@ -49,7 +49,7 @@ config() {
     mkdir -p ${HA_CONF}
 
     # TODO Add generation for rule engine to make rule engine schama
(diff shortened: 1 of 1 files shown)


saurabhkhanvilkar (approved):

conf/script/ha_setup
```diff
@@ -49,7 +49,7 @@ config() {
     mkdir -p ${HA_CONF}
 
     # TODO Add generation for rule engine to make rule engine schama
-    cp -rf ${SOURCE_HA_CONF}/rules_engine_schema_jbod.json ${HA_CONF}/${HA_CONF_RULE_FILE}
+    rule_engine_config
```
indrajitzagade, line 52: Currently we need to keep installing the empty rules engine file for the JBOD in main branch, are we changing that with this change?

Madhura-08: There was some confusion. I have fixed this now. Thanks !

indrajitzagade (approved):

ajaysrivas:
Please pass the DCO.

PR #207, opened 2020-10-15
Problem: s3auth dependency on motr is not justified

knekrasov:
## Problem Statement
<pre>
  <code>
  Story Ref (if any):
EOS-14300 "Remove dependency between s3auth and motr-ios-cN services"
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  No
  </code>
</pre>
## Problem Description
<pre>
  <code>
Currently s3auth clone is ordered after motr-ios-cN services.
This causes service restart when motr-ios-cN restarts. Noticed in EOS-14239.
s3auth resource is not dependent from motr system in any way, so this dependency can be considered for removal. Instead s3servers can be ordered after motr-ios-cN directly.
  </code>
</pre>
## Solution
<pre>
  <code>
1. Destroy constraint between s3auth and motr-io
2. Set similar dependency between s3server-cN-M and motr-io-cN (N=1,2)
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
N/A  </code>
</pre>

### DIFF
--- a/conf/script/build-ha-io
+++ b/conf/script/build-ha-io
@@ -719,8 +719,7 @@ s3auth_rsc_add() {
     sudo pcs -f $cib_file resource create s3auth systemd:s3authserver clone op \
         monitor interval=30
     sudo pcs -f $cib_file constraint order ldap-clone then s3auth-clone
-    sudo pcs -f $cib_file constraint order motr-ios-c1 then s3auth-clone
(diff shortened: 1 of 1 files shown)


azheregelya:
Pull request is relevant for R1 only. It may be reopened if such change will be required.

PR #208, opened 2020-10-16
[EOS-14191] To reduce failover/failback time kill m0d processes

VinothVijayakumar-dev:
- updating m0d@service using kill -9 from build-ha-io

Signed-off-by: Vinoth <vinoth.v@seagate.com>

## Problem Statement
<pre>
  <code>
  Story Ref (if any):
    To reduce failover/failback time kill m0d processes
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  Tested on HW with the changes on m0d@service
  </code>
</pre>

### DIFF
--- a/conf/script/build-ha-io
+++ b/conf/script/build-ha-io
@@ -621,8 +621,10 @@ motr_systemd_update() {
     log "${FUNCNAME[0]}: Adding Motr to Pacemaker..."
 
     cmd='
-    sudo sed "s/TimeoutStopSec=.*/TimeoutStopSec=5sec/"
-        -i /usr/lib/systemd/system/m0d@.service &&
+    sudo sed -e "s/TimeoutStopSec=.*/TimeoutStopSec=5sec/"
+             -e "/ExecStart=/aExecStop=/bin/kill -9 $MAIN_PID" \
+             -e "/ExecStart=/aExecStop=/usr/bin/echo 'shutting down process : ${MAINPID}'" \
+             -i /usr/lib/systemd/system/m0d@.service &&
     sudo systemctl daemon-reload'
     run_on_both $cmd
 }
--- a/conf/script/build-ha-io
+++ b/conf/script/build-ha-io
@@ -620,12 +620,14 @@ hax_rsc_add() {
 motr_systemd_update() {
     log "${FUNCNAME[0]}: Adding Motr to Pacemaker..."
 
-    cmd='
-    sudo sed -e "s/TimeoutStopSec=.*/TimeoutStopSec=5sec/"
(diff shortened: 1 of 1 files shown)


conf/script/build-ha-io
```diff
@@ -621,8 +621,10 @@ motr_systemd_update() {
     log "${FUNCNAME[0]}: Adding Motr to Pacemaker..."
 
     cmd='
-    sudo sed "s/TimeoutStopSec=.*/TimeoutStopSec=5sec/"
-        -i /usr/lib/systemd/system/m0d@.service &&
+    sudo sed -e "s/TimeoutStopSec=.*/TimeoutStopSec=5sec/"
```
mssawant, line 624: If we are killing the process anyway then I think we don't need the timeout change. And timeout can be specified in m0d@.service systemd unit file itself.

madhavemuri: Already timeout of 5min is there in m0d@service, let's keep 5 sec here.
In future  if we want to remove pkill, we will be aware of it.
May be let's add a comment here.
Also is such changes being applied during updates ?

mssawant: Yeah I think, let's add a comment saying
"Killing m0d to reduce stop time in case of failover, thus overriding the systemd timeout."
>Also is such changes being applied during updates ?

Yes, they are applied during update and restore too,
```
 [motr_systemd_update]='bootstrap_update_restore'
```

madhavemuri:
@mssawant : I think hax too can do pkill, which can reduce failover/failback time, do you see any issue with that ?

mssawant:
>@mssawant : I think hax too can do pkill, which can reduce failover/failback time, do you see any issue with that ?
 
Need to check, but if we are within the required failover/failback timings then let's not push it to avoid any regressions. Let's do such changes only if it is required.

conf/script/build-ha-io
madhavemuri: @VinothVijayakumar-dev : Can you add the comment here ?

mssawant (approved):

mssawant:
@VinothVijayakumar-dev  please test this change with deployment, update, failover, failback cases.

madhavemuri (approved):

VinothVijayakumar-dev:
I have tested this modification with deployment. shall update the failover and failback status.

indrajitzagade (approved):

VinothVijayakumar-dev:
closing this PR as this change required  in cortx-1.0 and in main branch
another https://github.com/Seagate/cortx-ha/pull/214 PR to merge this change in cortx-1.0

PR #215, opened 2020-10-19
Problem: Motr clients start before Motr IOS process starts functional

rajanikantchirmade:
Problem: Motr clients start before Motr IOS process starts functional

Pacemaker starts Motr clients after Motr servers processes, but since
pacemaker returns success after systemd process successfully starts.
In this case Motr server may not be started fully and functional to
cater requests and accepts requests from Motr client.

When Motr process starts and functional, it sends PROCESS_STARTED
HA event to Hax and that state is persisted in consul kv.

Solution:
Add check for PROCESS_STARTED HA state at Motr systemd unit file using
ExecStartPost script. This will make sure Pacemaker starts Motr clients
after Motr servers fully started and functional.

Ref: EOS-13872

## Problem Statement
<pre>
  <code>
  Story Ref (if any):
    Your Problem statement here...
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  Yes/No
  </code>
</pre>
## Problem Description
<pre>
  <code>
    Your Problem decription here...
  </code>
</pre>
## Solution
<pre>
  <code>
    Your Problem solution here...
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
    Unit Testing details here...
  </code>
</pre>

### DIFF
--- a/conf/script/build-ha-io
+++ b/conf/script/build-ha-io
@@ -620,10 +620,11 @@ hax_rsc_add() {
 motr_systemd_update() {
     log "${FUNCNAME[0]}: Adding Motr to Pacemaker..."
 
-    cmd='
-    sudo sed "s/TimeoutStopSec=.*/TimeoutStopSec=5sec/"
-        -i /usr/lib/systemd/system/m0d@.service &&
-    sudo systemctl daemon-reload'
+    cmd="
+    sudo sed -e "s/TimeoutStopSec=.*/TimeoutStopSec=5sec/"
+             -e \"/ExecStart=/aExecStartPost=/bin/sh -c 'while [[ \$\(/opt/seagate/cortx/hare/libexec/get-process-state %i\) != M0_CONF_HA_PROCESS_STARTED ]]\; do sleep 1\; done'\"
+             -i /usr/lib/systemd/system/m0d@.service &&
+    sudo systemctl daemon-reload"
     run_on_both $cmd
 }
 


andriytk:
@rajanikantchirmade , could you please explain - why is this a problem in the 1st place?

rajanikantchirmade:
@andriytk  

1. Pacemaker starts IOS service, and then it successfully starts
2. Pacemaker starts S3 services

But IOS not yet started all internal services like (CAS/MD etc) and before that S3 servers trying to connect endpoint of those services and timeout some times.

andriytk:
"... S3 servers trying to connect endpoint of those services and timeout some times."

I see... And what happens then after they "timeout"?

conf/script/build-ha-io
```diff
@@ -620,10 +620,11 @@ hax_rsc_add() {
 motr_systemd_update() {
     log "${FUNCNAME[0]}: Adding Motr to Pacemaker..."
 
-    cmd='
-    sudo sed "s/TimeoutStopSec=.*/TimeoutStopSec=5sec/"
-        -i /usr/lib/systemd/system/m0d@.service &&
-    sudo systemctl daemon-reload'
+    cmd="
+    sudo sed -e "s/TimeoutStopSec=.*/TimeoutStopSec=5sec/"
+             -e \"/ExecStart=/aExecStartPost=/bin/sh -c 'while [[ \$\(/opt/seagate/cortx/hare/libexec/get-process-state %i\) != M0_CONF_HA_PROCESS_STARTED ]]\; do sleep 1\; done'\"
```
andriytk, line 625: There is more proper way of doing this - by calling sd_notify(). But it was not done on purpose. From motr/m0d.c code:

```c
        /*
         * From the systemd's point of view, service can be considered as
         * started when it can handle incoming connections, which is already
         * true before m0_cs_start() is called. otherwise, if sd_notify() is
         * called after m0_cs_start() it leads to a deadlock, because different
         * m0d instances will wait for each other forever.
         */
        rc = sd_notify(0, "READY=1");
```

rajanikantchirmade: @andriytk ,
Right, but if sd_notify with `READY=1` moved after `m0_cs_start()`, it will block stating all other Motr process till current service started.

Currently Hare start Motr process on a node in sequentially. 
m0_cs_start() will try to connect all other service endpoints which are not started yet.

andriytk: Right. So better to do this change in Motr rather than here. (If we are about to do it at all.) I think.

ajaysrivas:
This is not required. Consider the code changes if we get an issue from field.

PR #218, opened 2020-10-20
build-ha-update: archive consul-kv-dump.json files

andriytk:
This should help to debug the problems with missing data in Consul.

Kudos to @NeeravChoudhari for the idea.


conf/script/build-ha-update
```diff
@@ -83,7 +83,11 @@ reset_all() {
 }
 
 echo 'Exporting Consul KV...'
-$consul_bin/consul kv export > $hare_dir/consul-conf-exported.json
+consul_kv_dump=$hare_dir/consul-kv-dump.json
+if [[ -f $consul_kv_dump ]]; do
+    cat $consul_kv_dump | xz > $consul_kv_dump-$(date -Iseconds).xz
```
knekrasov, line 88: `date -Iseconds` produces string with colons, so there will be no way to download such a file to a Windows machine:

```sh
date -Iseconds
2020-10-20T08:20:21-0600
```

Are we ok with this behavior?

andriytk: Good point. Let me change it...

knekrasov (approved):

mssawant (approved):

ajaysrivas:
@andriytk I guess, you need this change in cortx-1.0 but the PR is raised against main?

andriytk:
Yes, will do it for cortx-1.0 also after it's landed to main.

ajaysrivas (approved):

PR #242, opened 2020-11-12
Problem: Failback not happens for csm_web since failure-timeout not set

rajanikantchirmade:
Failback was not successful since pacemaker remember last attempt
to start resource which was failed. It required to set failure-timeout
for resource. Pacemaker will able to attempt starting of failed
resource after failure-timeout.

Solution:
Add failure-timeout for csm_web.

Ref: EOS-15013

### DIFF
--- a/conf/script/build-ha-csm
+++ b/conf/script/build-ha-csm
@@ -170,9 +170,9 @@ csm_rsc_add() {
     echo 'Adding csm resources and constraints...'
 
     sudo pcs -f $cib_file resource create csm-agent systemd:csm_agent op \
-        monitor interval=30s
+        monitor interval=30s meta failure-timeout=60s
     sudo pcs -f $cib_file resource create csm-web systemd:csm_web op \
-        monitor interval=30s
+        monitor interval=30s meta failure-timeout=60s
 
     sudo pcs -f $cib_file resource group add \
         csm-kibana kibana-vip kibana csm-web csm-agent


conf/script/build-ha-csm
```diff
@@ -172,7 +172,7 @@ csm_rsc_add() {
     sudo pcs -f $cib_file resource create csm-agent systemd:csm_agent op \
         monitor interval=30s
     sudo pcs -f $cib_file resource create csm-web systemd:csm_web op \
-        monitor interval=30s
+        monitor interval=30s meta failure-timeout=60s
```
mssawant, line 175: @rajanikantchirmade , I think you need to add migration-threshold to it as well because this is a active/passive resource. Pacemaker will try to restart the service  migration-threshold times and then failover the resource then try again on the primary node.
Is this tested on hardware?

rajanikantchirmade: @mssawant  Yes, it is tested on H/W setup, where it got reproduced.

mssawant: Okay, you need to make sure that pacemaker fails to start csm service and it failsover. Then once pacemaker is able to start the service on earlier node then it csm should automatically failback.
You  can try this,
```
delete csm_agent systemd unit file (backup the original before deleting), then disable and enable csm_agent resource, pacemaker will try to start but will fail, then csm stack will failover. After failover, replace the systemd unit file and see if csm fails back automatically.
```

mssawant: Approving the patch, @rajanikantchirmade please confirm the case as mentioned above.

mssawant (approved):

mssawant:
@rajanikantchirmade, Please fix DCO.

conf/script/build-ha-csm
azheregelya: @mssawant you mentioned csm-agent, but it doesn't have any failure timeout. So, i presume it is about csm-web.
If we want to have the same behavior for csm-agent - need to update its meta attributes as well.

azheregelya (approved):

conf/script/build-ha-csm
rajanikantchirmade: @azheregelya  Yes, failure-time out needed for csm-agent too. For csm_web I already tested yesterday I am looking for H/W setup to check again failover/failback and for csm-agent too.

rajanikantchirmade: @azheregelya  Done, added failure timeout for csm_agent too.

@mssawant  migration-threshold has exception if  `start-failure-is-fatal=true` for cluster. It's default value is `true`.
So resource is always migrate in this case.

```
[root@inteln13-m07 ~]# pcs property list --all | grep start-failure-is-fatal
 start-failure-is-fatal: true
[root@inteln13-m07 ~]#
```

rajanikantchirmade:
@mssawant 

Tested failover/failback with failure-timeout set for both resources csm-web & csm-agent.

ajaysrivas (approved):

andriytk:
> ... since pacemaker remember last attempt to start resource which was failed

Do we know why does the "last attempt" fails to start?

PR #247, opened 2020-11-21
Problem: Pacemaker manages 3rd party software haproxy

mssawant:
## Problem Statement
<pre>
  <code>
Pacemaker manages 3rd party software haproxy
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  No
  </code>
</pre>
## Problem Description
<pre>
  <code>
Haproxy being a 3rd party software, it need not be managed by pacemaker
along with the Seagate cortx stack.
  </code>
</pre>
## Solution
<pre>
  <code>
    Do not configure haproxy as a pacemaker resource.
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
    None
  </code>
</pre>

### DIFF
--- a/conf/script/build-ha-io
+++ b/conf/script/build-ha-io
@@ -701,7 +701,6 @@ systemd_units_disable() {
     log "${FUNCNAME[0]}: Disabling systemd units..."
     units_to_disable=(
         elasticsearch
-        haproxy
         rabbitmq-server
         statsd
         slapd
@@ -743,16 +742,6 @@ statsd_rsc_add() {
     log "${FUNCNAME[0]}: Adding statsd to Pacemaker..."
     sudo pcs -f $cib_file resource create statsd systemd:statsd clone op monitor interval=30s
     sudo pcs -f $cib_file constraint order els-search-clone then statsd-clone
-}
-
-haproxy_rsc_add() {
-    log "${FUNCNAME[0]}: Adding haproxy to pacemaker..."
-    sudo pcs -f $cib_file resource create haproxy-c1 systemd:haproxy op monitor interval=30s meta failure-timeout=60s
-    sudo pcs -f $cib_file resource create haproxy-c2 systemd:haproxy op monitor interval=30s meta failure-timeout=60s
-    sudo pcs -f $cib_file constraint location haproxy-c1 prefers $lnode=INFINITY
-    sudo pcs -f $cib_file constraint location haproxy-c1 avoids $rnode=INFINITY
-    sudo pcs -f $cib_file constraint location haproxy-c2 prefers $rnode=INFINITY
-    sudo pcs -f $cib_file constraint location haproxy-c2 avoids $lnode=INFINITY
 }
 
 get_s3_svc() {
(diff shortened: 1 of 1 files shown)


ajaysrivas:
@mssawant We need to remove other 3rd party components. Are we taking care of those in other PRs?

azheregelya (approved):
I see no other traces of haproxy. Looks good to me.

azheregelya:
> We need to remove other 3rd party components. Are we taking care of those in other PRs?

@ajaysrivas we own only haproxy pacemaker resource not haproxy itself. Since configuration and provisioning are done by other components I suppose we are not responsible for creating PRs, but probably shall participate in review.

cc: @mssawant @NeeravChoudhari

ajaysrivas:
@mssawant I guess, this is going to break update. Is this a case for build rebase? @NeeravChoudhari 
Code changes look fine but I will wait for results for the tests. We should at least install the build and check if all the services are starting fine.

mssawant:
@ajaysrivas , @azheregelya , updated patch to handle update. Added a check to delete the resource if it exists only.

conf/script/prov-ha-reset
```diff
@@ -54,7 +54,9 @@ resources=(
     ip-c{2,1}
 )
 for r in ${resources[@]}; do
-    pcs resource delete $r || true
+    if [[ $(pcs resource show $r 2> /dev/null) ]]; then
+        pcs resource delete $r || true
```
azheregelya, line 58: `|| true` in the end makes `if .. then` check redundant since `pcs resource delete` failure will be ignored anyway.
I would suggest to keep this file as is.

mssawant: Yes, we need not add a check, just thought of still having it just in case if the error is parsed by some external script like provisioner. May be then let's just re-direct the error to /dev/null.

mssawant: Actually you are right, let's just keep it as the original.

mssawant: Done.

indrajitzagade (approved):

azheregelya:
Extraction from secret chat dedicated to haproxy removal:

I checked generated haproxy.cfg on alive cluster. Lets check s3servers backend.
On example below we see 1 problem:
Only 10 s3servers which belong to current node are specified. Since customer-managed load balancing will know nothing about s3servers, haproxy have to reroute I/O requests to alive s3servers. Those alive s3servers may be located only on other (neighbor) node. But they are missing in this list.
So, we have one more action item to do: provisioner component has to add all s3server instances into haproxy.cfg with corresponding IP addresses. Usage of statically configured private IP addresses is correct here from my point of view.
```
    server s3-instance-100 192.168.0.1:28081 check  # s3 instance 100
    server s3-instance-101 192.168.0.1:28082 check  # s3 instance 101
    server s3-instance-102 192.168.0.1:28083 check  # s3 instance 102
    server s3-instance-103 192.168.0.1:28084 check  # s3 instance 103
    server s3-instance-104 192.168.0.1:28085 check  # s3 instance 104
    server s3-instance-105 192.168.0.1:28086 check  # s3 instance 105
    server s3-instance-106 192.168.0.1:28087 check  # s3 instance 106
    server s3-instance-107 192.168.0.1:28088 check  # s3 instance 107
    server s3-instance-108 192.168.0.1:28089 check  # s3 instance 108
    server s3-instance-109 192.168.0.1:28090 check  # s3 instance 109
    server s3-instance-110 192.168.0.1:28091 check  # s3 instance 110
```

ajaysrivas:
This is not required. We will not remove haproxy from pacemaker.

PR #255, opened 2020-11-27
EOS-14644: don't stop LNet and Motr on failover

chumakd:
## Problem Statement
<pre>
  <code>
Pacemaker cluster failover takes more that 60 seconds.
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  Yes
  </code>
</pre>
## Problem Description
<pre>
  <code>
As part of failover and failback transitions, Pacemaker stops native
Lnet resources on the healthy srvnode, which implies stopping Motr,
S3server and Hare resources as well. And then starts them again. This
extra stop/start cycle can now be avoided since EOS-6023 has been fixed.
  </code>
</pre>
## Solution
<pre>
  <code>
Pacemaker's resource configuration needs to be updated so that native
node resources continue to work and don't get restarted during failover
and failback.  
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
Tested on inteln13/14, failover time reduced to ~30sec. Failback time is under 60sec.
  </code>
</pre>

### DIFF
--- a/conf/script/build-ha-io
+++ b/conf/script/build-ha-io
@@ -535,19 +535,7 @@ consul_rsc_add() {
 motr_kernel_rsc_add() {
     log "${FUNCNAME[0]}: Adding Motr kernel module to Pacemaker..."
     sudo pcs -f $cib_file resource create motr-kernel systemd:motr-kernel clone
-    sudo pcs -f $cib_file constraint order lnet-c1 then motr-kernel-clone
-    sudo pcs -f $cib_file constraint order lnet-c2 then motr-kernel-clone
-
-    # Make sure motr-kernel is always stopped before lnet-c1/2
-    # is started. Otherwise, some transition abort in between may
-    # prevent motr-kernel restart. This is a workaround suggested
-    # at https://bugs.clusterlabs.org/show_bug.cgi?id=5428#c5.
-    sudo pcs -f $cib_file constraint order stop motr-kernel-clone then \
-                                           start lnet-c1 \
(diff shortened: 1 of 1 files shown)


conf/script/build-ha-io
```diff
@@ -535,19 +535,7 @@ consul_rsc_add() {
...
-    # is started. Otherwise, some transition abort in between may
-    # prevent motr-kernel restart. This is a workaround suggested
-    # at https://bugs.clusterlabs.org/show_bug.cgi?id=5428#c5.
-    sudo pcs -f $cib_file constraint order stop motr-kernel-clone then \
-                                           start lnet-c1 \
-                                           kind=Optional symmetrical=false
-    sudo pcs -f $cib_file constraint order stop motr-kernel-clone then \
-                                           start lnet-c2 \
-                                           kind=Optional symmetrical=false
+    sudo pcs -f $cib_file constraint order lnet-clone then motr-kernel-clone
```
azheregelya, line 538: `lnet-clone` is used here instead of lnet-c1. I would expect also changes where `lnet-clone` resource is created (or changed to clone), but don't see it.

azheregelya (changes requested):
Looks like there is a bug here. Please take a look?

azheregelya (approved):

indrajitzagade (approved):

chumakd:
I think this change is not relevant for R2, so it's probably better to close this PR for now w/o merging. It can be re-opened later if needed.

PR #259, opened 2020-12-04
EOS-15329: Move cluster into standby mode before deleting the resources 

VijayThakkar132:
Signed-off-by: Vijay Thakkar <vijaykumar.thakkar@seagate.com>

## Problem Statement
<pre>
  <code>
  Story Ref (if any):
    EOS-15329: Build 515: Unsuccessful update to build 529 after primary node replacement
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  Yes/No
  </code>
</pre>
## Problem Description
<pre>
  <code>
    Observed build-ha-update script took more time to delete and recreate resources.
  </code>
</pre>
## Solution
<pre>
  <code>
    Did changes to move cluster into standby mode before deleting resources and moving back to unstandby mode after recreating resources again.
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
    Unit Testing details here...
  </code>
</pre>

### DIFF
--- a/conf/script/build-ha-update
+++ b/conf/script/build-ha-update
@@ -107,7 +107,6 @@ sudo pcs cluster standby --all
 reset_all
 
 sudo pcs cluster unstandby --all
-sleep 2
 sudo pcs resource cleanup
 
 sudo pcs cluster cib $cib_file
--- a/conf/script/build-ha-update
+++ b/conf/script/build-ha-update
@@ -102,12 +102,12 @@ if [[ -f $consul_kv_dump ]]; then
 fi
 $consul_bin/consul kv export > $consul_kv_dump
 
+# Move cluster into standby mode before deleting the resources
 sudo pcs cluster standby --all
 
 reset_all
 
 sudo pcs cluster unstandby --all
-sudo pcs resource cleanup
 
 sudo pcs cluster cib $cib_file
 
--- a/conf/script/build-ha-update
+++ b/conf/script/build-ha-update
@@ -103,7 +103,8 @@ fi
 $consul_bin/consul kv export > $consul_kv_dump
 
 # Move cluster into standby mode before deleting the resources
-sudo pcs cluster standby --all
+pcs_status=$(sudo pcs status)
+echo $pcs_status | grep "Node srvnode-1: standby" && echo $pcs_status | grep "Node srvnode-2: standby" || sudo pcs cluster standby --all
 
 reset_all
 
--- a/conf/script/build-ha-update
+++ b/conf/script/build-ha-update
@@ -111,6 +111,11 @@ $consul_bin/consul kv export > $consul_kv_dump
 sudo pcs cluster standby --all
 
 reset_all
+
+sudo pcs cluster unstandby --all
+sleep 2
+sudo pcs resource cleanup
+
 sudo pcs cluster cib $cib_file
 
 echo 'Updating ha resources for IO path, SSPL, CSM and UDS...'
@@ -130,10 +135,6 @@ sudo pcs cluster cib-push $cib_file --config
 
 $ha_script/build-cortx-ha init
 
-sudo pcs resource cleanup
-
-sudo pcs cluster unstandby --all
-sleep 2
 sudo pcs resource cleanup
 
 echo 'Importing Consul KV...'
--- a/conf/script/build-ha-update
+++ b/conf/script/build-ha-update
@@ -112,7 +112,7 @@ sudo pcs cluster unstandby --all
 
 while ! sudo pcs status | grep "Online: \[ srvnode-1 srvnode-2 ]"
 do
-    echo 'Waiting for cluster becomes online'
+    echo 'Waiting till cluster becomes online'
     sleep 1
 done
 
--- a/conf/script/build-ha-update
+++ b/conf/script/build-ha-update
@@ -14,10 +14,16 @@
 # with this program. If not, see <https://www.gnu.org/licenses/>. For any questions
 # about this software or licensing, please email opensource@seagate.com or
 # cortx-questions@seagate.com.
+HA_LOG="/var/log/seagate/cortx/ha"
 
 set -eu -o pipefail
 export PS4='+ [${BASH_SOURCE[0]##*/}:${LINENO}${FUNCNAME[0]:+:${FUNCNAME[0]}}] '
 # set -x
+
+# Add Log
+mkdir -p "${HA_LOG}"/
+exec &>> "${HA_LOG}"/${0##*/}.log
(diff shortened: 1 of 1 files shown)


indrajitzagade:
> 
> 
> Signed-off-by: Vijay Thakkar [vijaykumar.thakkar@seagate.com](mailto:vijaykumar.thakkar@seagate.com)
> ## Problem Statement
> 
>   ```
> 
>   Story Ref (if any):
>     EOS-15329: Build 515: Unsuccessful update to build 529 after primary node replacement
>   
> ```
> 
> 
> ## Unit testing on RPM done
> 
>   ```
> 
>   Yes/No
>   
> ```
> 
> 
> ## Problem Description
> 
>   ```
> 
>     Observed build-ha-update script took more time to delete and recreate resources.
>   
> ```
> 
> 
> ## Solution
> 
>   ```
> 
>     Did changes to move cluster into standby mode before deleting resources and moving back to unstandby mode after recreating resources again.
>   
> ```
> 
> 
> ## Unit Test Cases
> 
>   ```
> 
>     Unit Testing details here...
>   
> ```

@VijayThakkar132 Please share how this patch was tested and the test observations.

VijayThakkar132:
@indrajitzagade Testing is not yet completed. I will run build-ha-update with and with-out changes and will measure execution time to verify improvement after moving the cluster into standby mode while deleting and recreating resources.

VijayThakkar132:
Hi @indrajitzagade,

@ajaysrivas has suggested the below steps to verify the bug is resolved after the suggested solution for the final testing. Please also share your thoughts on how can we verify the solution as it's not reproducible every time.

1. Generate a custom build with the fix.
2. Kill some process which has a timeout(e.g sspl) to move cluster into an unstable state
3. Perform software upgrade parallelly to custom build and verify the reported issue is reproducible or not?
4. Repeat the above steps multiple times

indrajitzagade:
> 
> 
> Hi @indrajitzagade,
> 
> @ajaysrivas has suggested the below steps to verify the bug is resolved after the suggested solution for the final testing. Please also share your thoughts on how can we verify the solution as it's not reproducible every time.
> 
>     1. Generate a custom build with the fix.
> 
>     2. Kill some process which has a timeout(e.g sspl) to move cluster into an unstable state
> 
>     3. Perform software upgrade parallelly to custom build and verify the reported issue is reproducible or not?
> 
>     4. Repeat the above steps multiple times

@VijayThakkar132 I think these steps should help reproduce the issue and verify the fix.

VijayThakkar132:
Tried multiple times to reproduce the reported issue without and with a fix but somehow it was not reproducible. I observed that there is an improvement in the execution time of the build-ha-update script.

indrajitzagade:
> 
> 
> Tried multiple times to reproduce the reported issue without and with a fix but somehow it was not reproducible. I observed that there is an improvement in the execution time of the build-ha-update script.

@VijayThakkar132 The software update worked fine with this patch, rt?

VijayThakkar132:
> > Tried multiple times to reproduce the reported issue without and with a fix but somehow it was not reproducible. I observed that there is an improvement in the execution time of the build-ha-update script.
> 
> @VijayThakkar132 The software update worked fine with this patch, rt?

Yes, @indrajitzagade It worked fine

conf/script/build-ha-update
```diff
@@ -89,7 +89,14 @@ if [[ -f $consul_kv_dump ]]; then
 fi
 $consul_bin/consul kv export > $consul_kv_dump
 
+sudo pcs cluster standby --all
+
 reset_all
+
+sudo pcs cluster unstandby --all
+sleep 2
```
indrajitzagade, line 97: Why we need sleep here? Is 2 seconds sufficient?

VijayThakkar132: Added delay to give some time to resources to start successfully before doing the resource cleanup. I tested and went well. So, I feel 2 secs delay is sufficient.

indrajitzagade: We should be careful with adding delays in the software update path. Already we had tasks to optimize the software update time. I would recommend removing this delay if it is not required absolutely.

VijayThakkar132: Sure. I will give a try without delay and update the results.

VijayThakkar132: It worked fine without delay also. So, removing delay in the next patch as it's not required.

conf/script/build-ha-update
```diff
@@ -89,7 +89,13 @@ if [[ -f $consul_kv_dump ]]; then
 fi
 $consul_bin/consul kv export > $consul_kv_dump
 
+sudo pcs cluster standby --all
+
 reset_all
+
+sudo pcs cluster unstandby --all
+sudo pcs resource cleanup
```
indrajitzagade, line 97: Do we really need this command? All the resources would have been deleted when we are here, rt?

indrajitzagade: > 
> 
> @VijayThakkar132 
 Do we really need this command? All the resources would have been deleted when we are here, rt?

VijayThakkar132: Yes, I think not needed. Let me give a try on the hardware setup to verify how it goes without cleanup.

VijayThakkar132: Software update went well without cleanup also. So, removing cluster cleanup in the next patch.

ajay-paratmandali (approved):

conf/script/build-ha-update
```diff
@@ -102,7 +102,13 @@ if [[ -f $consul_kv_dump ]]; then
 fi
 $consul_bin/consul kv export > $consul_kv_dump
 
+# Move cluster into standby mode before deleting the resources
+sudo pcs cluster standby --all
```
ajaysrivas, line 106: Is it possible to put a check here if the cluster is already in standby then do not call standby command?

conf/script/build-ha-update
```diff
@@ -102,7 +102,13 @@ if [[ -f $consul_kv_dump ]]; then
...
+
 reset_all
+
+sudo pcs cluster unstandby --all
```
ajaysrivas, line 110: A check here again, proceed only when cluster has become active?

conf/script/build-ha-update
VijayThakkar132: Yes, Added check to skip command if the cluster is already in standby mode

conf/script/build-ha-update
VijayThakkar132: done

ajaysrivas (approved):

indrajitzagade (approved):

PR #260, opened 2020-12-04
changes for EOS-15135 Add logging to build-ha-* scripts 

ArchanaLimaye:
## Problem Statement
<pre>
  <code>
    Logging is missing for the HA build and reset scripts 
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  No
  </code>
</pre>
## Problem Description
<pre>
  <code>
    Logging is missing for the HA build and reset scripts 
  </code>
</pre>
## Solution
<pre>
  <code>
    Logging added to the files 
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
    For testing, custom build will be created with the changes and will be deployed on hardware 
  </code>
</pre>

### DIFF
--- a/conf/script/build-cortx-ha
+++ b/conf/script/build-cortx-ha
@@ -20,7 +20,17 @@ set -eu -o pipefail
 
 HA_PATH="<HA_PATH>"
 
+log_file=${BUILD_CORTX_HA_LOG_FILE:-/var/log/build-cortx-ha.log}
+exec &>> $log_file
+exec &> >(stdbuf -oL gawk '{ print strftime("%Y-%m-%d %H:%M:%S"), $0 }')
+
+log() {
+    logger --stderr --tag build-cortx-ha "$*"
+}
+
+
 init() {
+    log "${FUNCNAME[0]}: Init cluster"
     # Get node name
     node_list=$(salt-call --local pillar.get cluster:node_list --output=json)
     host1=$(echo $node_list | jq '.["local"][0]' | sed s/\"//g)
@@ -78,6 +88,7 @@ init() {
 
 cleanup() {
     echo "Delete resources"
+    log "${FUNCNAME[0]}: Delete resources"
     resources=(
         io_path_health-c1
         io_path_health-c2
--- a/conf/script/build-cortx-ha
+++ b/conf/script/build-cortx-ha
@@ -96,6 +96,7 @@ cleanup() {
         sspl_primary_hw
     )
     for r in ${resources[@]}; do
+        log "${FUNCNAME[0]}: Deleting resource: $r"
         pcs resource delete $r || true
     done
 }
--- a/conf/script/build-cortx-ha
+++ b/conf/script/build-cortx-ha
@@ -20,13 +20,17 @@ set -eu -o pipefail
 
 HA_PATH="<HA_PATH>"
 
-log_file=${BUILD_CORTX_HA_LOG_FILE:-/var/log/build-cortx-ha.log}
+log_file=${BUILD_CORTX_HA_LOG_FILE:-/var/log/cortx-ha.log}
 exec &>> $log_file
 exec &> >(stdbuf -oL gawk '{ print strftime("%Y-%m-%d %H:%M:%S"), $0 }')
 
 log() {
     logger --stderr --tag build-cortx-ha "$*"
 }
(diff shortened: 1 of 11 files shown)


conf/script/build-cortx-ha
```diff
@@ -20,7 +20,17 @@ set -eu -o pipefail
 
 HA_PATH="<HA_PATH>"
 
+log_file=${BUILD_CORTX_HA_LOG_FILE:-/var/log/build-cortx-ha.log}
```
ajay-paratmandali, line 23: `log_file=${BUILD_CORTX_HA_LOG_FILE:-/var/log/build-cortx-ha.log}`

Here we are creating new file for each script which will create too many log file for same task.

can we use single generic file like `ha_resource_setup.log` single file for all script

Also add this to initial of file to identify script log
```
log "****************************************************************"
log "${0##*/}"
log "****************************************************************"
```

ArchanaLimaye: The log file name is already indicating the script file for which this log is created. 
Will adding the same name in the log file help in any additional way ?

ArchanaLimaye: Regarding 
>>can we use single generic file like ha_resource_setup.log single file for all script
we are already adding all the logs to the syslog by using the logger utility. So they are available in a single file there. 
I though that having separate files will help in locating logs from individual scripts.

However I am fine with having a single file if that is going to give better usability

conf/script/build-cortx-ha
```diff
@@ -20,7 +20,17 @@ set -eu -o pipefail
...
+exec &>> $log_file
+exec &> >(stdbuf -oL gawk '{ print strftime("%Y-%m-%d %H:%M:%S"), $0 }')
+
+log() {
+    logger --stderr --tag build-cortx-ha "$*"
+}
+
+
 init() {
+    log "${FUNCNAME[0]}: Init cluster"
```
indrajitzagade, line 33: How do we control logging level?

conf/script/build-ha-io
```diff
@@ -131,7 +136,7 @@ cdf=${1:-}
 argsfile=${2:-}
 
 hare_dir=/var/lib/hare
-log_file=${LOG_FILE:-/var/log/build-ha.log}
+#log_file=${LOG_FILE:-/var/log/build-ha.log}
```
indrajitzagade, line 139: Please delete before final merge.

conf/script/build-cortx-ha
azheregelya: It is deployment stage. I think logging is not supposed to be controlled somehow during provisioning.

azheregelya: It is a good point for low-level HA requirements discussion. Runtime logging shall be controllable for sure. But deployment - i see not use case in that so far.

conf/script/build-cortx-ha
```diff
@@ -78,6 +88,7 @@ init() {
 
 cleanup() {
     echo "Delete resources"
+    log "${FUNCNAME[0]}: Delete resources"
     resources=(
```
azheregelya, line 92: There is `for` loop below where each resource is being deleted. Please add log message to indicate which exactly resource is being deleted at this moment. This makes easier to understand on which resource script could fail.

azheregelya (changes requested):

conf/script/build-ha-io
ArchanaLimaye: Yes will remove before merge

conf/script/build-cortx-ha
ArchanaLimaye: okay

conf/script/build-cortx-ha
ArchanaLimaye: This log  is added for the sake of completeness; so that every stage gets reported in the logs. 
We can remove it, if that is the general consensus

we are currently using the linux logger utility for logging here. 
I have been unable to find any way to configure logging level using the linux logger. 
>>How do we control logging level?
So not sure how this can be achieved even after deployment stage.

indrajitzagade: ok, understood.

indrajitzagade (approved):

azheregelya (approved):

PR #292, opened 2021-02-01
EOS-16878: SSPL resource agent improvements

VijayThakkar132:
Signed-off-by: Vijay Thakkar <vijaykumar.thakkar@seagate.com>

## Problem Statement
<pre>
  <code>
  Story Ref (if any):
    EOS-16878: HA (pcs) did not detect that SSPL had crashed and did not restart it
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  Yes
  </code>
</pre>
## Problem Description
<pre>
  <code>
    Observed that demoted node is taking too much time after the forcefully kill.
  </code>
</pre>
## Solution
<pre>
  <code>
    Added code changes to remove state file in the stop method and also added couple of improvement
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
    Tested changes in the live setup and observed that sspl service starts immediately after the demotion.
  </code>
</pre>

### DIFF
--- a/ha/resource/sspl
+++ b/ha/resource/sspl
@@ -144,10 +144,8 @@ stateful_start() {
         start_sspl_service
         if [ $? == 0 ]; then
             return $OCF_RUNNING_MASTER
-        elif [ $? == 1 ]; then
-            return $OCF_FAILED_MASTER
         else
-            return $OCF_ERR_GENERIC
+            return $OCF_FAILED_MASTER
         fi
     fi
     stateful_update "slave"
--- a/ha/resource/sspl
+++ b/ha/resource/sspl
@@ -27,7 +27,7 @@
 #	SSPL OCF Stateful master/slave RA.
 
 # Return codes and its menaning
-# 0: OCF_SUCESS: Successful Execution
+# 0: OCF_SUCCESS: Successful Execution
 # 9: OCF_MASTER_FAILED: Resource is failed in Master mode
 # 7: OCF_NOT_RUNNING: Resource is safely stopped.
 
@@ -121,6 +121,10 @@ stateful_start() {
         # CRM Error - Should never happen
         log "START (master)..."
         systemctl start sspl-ll
+        while ! systemctl is-active sspl-ll | grep "active"
+        do
+            sleep 1
+        done
         log "START (master): $?"
         return $OCF_RUNNING_MASTER
     fi
@@ -128,6 +132,10 @@ stateful_start() {
     "${HA_SBIN_DIR}/crm_master" -l reboot -v 1
     log "START... (crm_reboot: $?)"
     systemctl start sspl-ll
+    while ! systemctl is-active sspl-ll | grep "active"
+    do
+        sleep 1
+    done
     log "START: $?"
 
     return $OCF_SUCCESS
@@ -177,7 +185,8 @@ stateful_stop() {
     log "STOP..."
     systemctl stop sspl-ll
     log "STOP: $?"
-    return $OCF_SUCESS
+    rm -rf "${OCF_RESKEY_state}"
+    return $OCF_SUCCESS
 }
 
 stateful_monitor() {
--- a/ha/resource/sspl
+++ b/ha/resource/sspl
@@ -115,30 +115,45 @@ stateful_check_state() {
     return 7
 }
 
+start_sspl_service() {
+        systemctl start sspl-ll
+        echo "START (master): $?"
+        while true
+        do
+            status=`systemctl is-active sspl-ll`
+            if [[ $status == "active" ]]; then
+                return 0
+            elif [[ $status == "failed" ]]; then
+                return 1
+            else
+                sleep 1
+            fi
+        done
+}
+
 stateful_start() {
     stateful_check_state "master"
     if [ $? -eq 0 ]; then
         # CRM Error - Should never happen
         log "START (master)..."
-        systemctl start sspl-ll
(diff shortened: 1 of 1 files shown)


Madhura-08 (approved):

ha/resource/sspl
```diff
@@ -121,13 +121,21 @@ stateful_start() {
         # CRM Error - Should never happen
         log "START (master)..."
         systemctl start sspl-ll
+        while ! systemctl is-active sspl-ll | grep "active"
```
ajaysrivas, line 124: you should be checking "failed" status as well and if service fails to start, return proper OCF code.
Applies to code below too. You can add another function as code will be common and call the function from both the places.

ajaysrivas (changes requested):

ha/resource/sspl
VijayThakkar132: Addressed comment in the latest patch

ha/resource/sspl
```diff
@@ -115,22 +115,45 @@ stateful_check_state() {
...
 
+start_sspl_service() {
+        systemctl start sspl-ll
+        echo "START (master): $?"
+        while true
+        do
+            status=`systemctl is-active sspl-ll`
+            if [[ $status == "active" ]]; then
+                return 0
+            elif [[ $status == "failed" ]]; then
```
ajaysrivas, line 126: what about "unknown" status?
How should that be handled?
@ajay-paratmandali, have you checked that in dynamicFid resource agent?

VijayThakkar132: @ajaysrivas What is the expected behavior when the current service state is "unknown"?  Do we need to wait till the service starts or returns immediately?

ajaysrivas: Not sure. This is the reason, I asked question :-)
Probably we should check if it remains "unknown" for some time (5 sec) then return 1.

VijayThakkar132: Yes, It's good to wait for 5 sec. Did the changes in the latest patch.

indrajitzagade (approved):

ha/resource/sspl
ajaysrivas: I meant, continuously for 5 sec. So the count should be set to zero in other if conditions. I guess, 5 sec is fine for a stable status, so this is okay too.

ha/resource/sspl
```diff
@@ -115,22 +115,50 @@ stateful_check_state() {
...
         log "START (master)..."
-        systemctl start sspl-ll
-        log "START (master): $?"
-        return $OCF_RUNNING_MASTER
+        start_sspl_service
+        if [ $? == 0 ]; then
+            return $OCF_RUNNING_MASTER
+        elif [ $? == 1 ]; then
+            return $OCF_FAILED_MASTER
+        else
```
ajaysrivas, line 149: return code is either 0 or 1, it will never reach here.

VijayThakkar132: Yes, Not required elif condition. Updated in the latest patch

ajaysrivas (approved):

ha/resource/sspl
```diff
@@ -115,22 +115,48 @@ stateful_check_state() {
...
+                return 1
+            elif [[ $status == "unknown" ]]; then
+                ((count++))
+                if [ $count == 5 ]; then
+                    return 1
+                fi
+            fi
+            sleep 1
+        done
+}
```
ujjwalpl, line 137: No return value at the end. Caller is looking for a return value.

ha/resource/sspl
```diff
@@ -115,22 +115,48 @@ stateful_check_state() {
     return 7
 }
 
+start_sspl_service() {
+        systemctl start sspl-ll
+        echo "START (master): $?"
```
ujjwalpl, line 120: Code should be written this way for better readability ... 

```
start_sspl_service() {
	echo "START (master): $?"
	systemctl start sspl-ll
    
    # check status
    for retry in {1..5}; do 
    	status=$(systemctl is-active sspl-ll)
    	case $status in
    		active )
    			return 0
    			;;
    			
    		failed )
    			return 1
    			;;
    			
    		* )
    			;;
    	esac
    	sleep 1
    done
    return 1
}
```

ha/resource/sspl
```diff
@@ -115,22 +115,48 @@ stateful_check_state() {
...
+
 stateful_start() {
     stateful_check_state "master"
     if [ $? -eq 0 ]; then
         # CRM Error - Should never happen
         log "START (master)..."
-        systemctl start sspl-ll
-        log "START (master): $?"
-        return $OCF_RUNNING_MASTER
+        start_sspl_service
```
ujjwalpl, line 144:
```
start_sspl_service && return $OCF_RUNNING_MASTER
return $OCF_FAILED_MASTER
```

ha/resource/sspl
VijayThakkar132: Raised below PR to address the above comment.
https://github.com/Seagate/cortx-ha/pull/296

ha/resource/sspl
VijayThakkar132: Raised below PR to address the above comment.
https://github.com/Seagate/cortx-ha/pull/296

ha/resource/sspl
VijayThakkar132: Raised below PR to address the above comment.
https://github.com/Seagate/cortx-ha/pull/296

PR #352, opened 2021-03-25
EOS-18593: Code changes for first and subsequent node addition to cluster

indrajitzagade:
## Problem Statement
<pre>
  <code>
  Story Ref (if any):
    Your Problem statement here...
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  Yes/No
  </code>
</pre>
## Problem Description
<pre>
  <code>
    Add support for first and subsequent node addition to cluster
  </code>
</pre>
## Solution
<pre>
  <code>
    Updated mini-provisioner to support first and subsequent node addition to cluster
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
    Tested with two nodes. Ran mini-provisioner post_install, init and config commands on first node and then on second node. The cluster was created and two nodes were seen in the cluster. Also, tried cleanup was tried on both nodes.
  </code>
</pre>


ha/setup/ha_setup.py
```diff
@@ -194,30 +196,79 @@ def process(self):
...
+        cluster_exists = self._confstore.key_exists(const.CLUSTER_CONFSTORE_NODES_KEY)
+        Log.info(f"Cluster exists? {cluster_exists}")
+        if cluster_exists == False:
+            try:
+                # Create cluster
+                Log.info(f"Creating cluster: {cluster_name} with node: {node_name}")
+                cluster_output: str = self._cluster_manager.cluster_controller.create_cluster(
+                    cluster_name, cluster_user, cluster_secret, node_name)
+                Log.info(f"Cluster creation output: {cluster_output}")
+                # TODO: Handle race condition cluster and resource create with global check.
```
ajaysrivas, line 209: which race condition?

ha/setup/ha_setup.py
```diff
@@ -353,9 +404,33 @@ def process(self):
...
+                        Log.info(f"Removed {node_name} from the cluster")
+                        node_removed = True
+                        break
+                    except Exception as e:
+                        Log.error(f"Removing {node_name} using remote node {remote_node} failed with error: {e}, \
+                                if available, will try with other node")
+
+            if node_removed == False:
+                Log.info(f"{node_name} remove failed/last node in the cluster, destroy the cluster")
+                output = self._execute.run_cmd(const.PCS_CLUSTER_DESTROY, check_error=True)
```
ajaysrivas, line 429: Using node_removed flag for identifying the last node. I think, it will cause the problem if "remove node" fails for an intermediate node. We can identify the last node by count number of nodes present in conf_store.

ajaysrivas:
A general comment - please verify that all the operations are idempotent. If the command is executed for a node again, should not fail and do nothing if it has already done what it is supposed to do.

ha/setup/ha_setup.py
indrajitzagade: This is an existing comment, @ajay-paratmandali can you please provide more info?

ha/setup/ha_setup.py
indrajitzagade: I think no. const.PCS_CLUSTER_DESTROY will destroy the cluster only on that specified node. The cluster on other nodes remains intact.
I think "remove node" is the best thing to try since it will remove the specified node from cluster. The "remove node" is to be tried from the other nodes in the cluster (similar to "add node"). So if somehow the "remove node" from all other node fails, then we try const.PCS_CLUSTER_DESTROY as a last resort. Also, if it is the last node being removed then we cannot do "remove node", hence do const.PCS_CLUSTER_DESTROY.

ha/setup/ha_setup.py
ajay-paratmandali: It was rase condition between

create_cluster()
create_resource()
But it got addressed by 

`self._confstore.set(f"{const.CLUSTER_CONFSTORE_NODES_KEY}/{node_name}")`
we can remove this TODO

indrajitzagade: Removed the TODO.

ha/setup/ha_setup.py
indrajitzagade: > 
> 
> A general comment - please verify that all the operations are idempotent. If the command is executed for a node again, should not fail and do nothing if it has already done what it is supposed to do.

@ajaysrivas I think we are good with post_install and init already. Added some checks in config and cleanup.

indrajitzagade:
> 
> 
> A general comment - please verify that all the operations are idempotent. If the command is executed for a node again, should not fail and do nothing if it has already done what it is supposed to do.

@ajaysrivas I think we are good with post_install and init already. Added some checks in config and cleanup.

ha/setup/ha_setup.py
```diff
@@ -196,31 +198,84 @@ def process(self):
...
-                # Put cluster in standby mode
-                standby_output: str = self._cluster_manager.node_controller.standby(node_name)
-                Log.info(f"Put node in standby output: {standby_output}")
-                if json.loads(standby_output).get("status") != STATUSES.FAILED.value:
-                    Log.info("Creating pacemaker resources")
-                    create_all_resources(s3_instances=s3_instances)
-                    Log.info("Created pacemaker resources successfully")
+        self._cluster_manager = CortxClusterManager()
+        # Check if the cluster exists already, if not, create one.
+        cluster_exists = self._confstore.key_exists(const.CLUSTER_CONFSTORE_NODES_KEY)
```
ajay-paratmandali, line 203: `# Check if the cluster exists already, if not, create one.`
Put this sentences in log.info
If consul down or halt log will be useful to track

ha/setup/ha_setup.py
```diff
@@ -196,31 +198,84 @@ def process(self):
...
+                Log.error(f"Cluster creation failed; destroying the cluster. Error: {e}")
+                output = self._execute.run_cmd(const.PCS_CLUSTER_DESTROY, check_error=True)
+                Log.info(f"Cluster destroyed. Output: {output}")
+                # Delete the node from nodelist if it was added in the store
+                if self._confstore.key_exists(f"{const.CLUSTER_CONFSTORE_NODES_KEY}/{node_name}"):
+                    self._confstore.delete(f"{const.CLUSTER_CONFSTORE_NODES_KEY}/{node_name}")
+                raise HaConfigException("Cluster creation failed")
+        else:
+            node_added = False
+            # If this node in cluster already.
```
ajay-paratmandali, line 236: Put in log
`# If this {node} node in cluster already`

ha/setup/ha_setup.py
indrajitzagade: done.

ha/setup/ha_setup.py
indrajitzagade: done

ajay-paratmandali (approved):

ajaysrivas (approved):

PR #353, opened 2021-03-25
EOS-17123: Cluster validation routines for ha_setup test

azheregelya:
## Problem Statement
<pre>
  <code>
  Story Ref (if any): https://jts.seagate.com/browse/EOS-17123
  Implement test command for cluster
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  In-Progress
  </code>
</pre>
## Problem Description
<pre>
  <code>
    Need to implement cluster validation routines to check that cluster built according to the plan and working fine.
  </code>
</pre>
## Solution
<pre>
  <code>
1. Create component json files with cluster structure.
2. Implement ClusterLayout class to parse component json files.
3. Implement ClusterStatus class to parse pacemaker status.
4. Implement to set of functions to compare ClusterStatus against expected ClusterLayout and do additional sanity checks.
5. Implement a dedicate CLI script run checks and fail with logs if cluster is not valid.
6. Put this code to standalone RPM package. (excluded and will be made in separate PR)
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
1. Unit-tests for ClusterLayout to make sure that component json files are parsed correctly
2. Unit-tests for ClusterStatus to make sure that all combinations for resource configurations can be parsed and represented by class objects.
3. Unit-tests to check that validation works as expected.
4. CLI work are expected to be tested manually.
  </code>
</pre>

### DIFF
--- a/conf/etc/v2/components/csm.json
+++ b/conf/etc/v2/components/csm.json
@@ -1,0 +1,161 @@
+{
+    "csm-agent": {
+        "parameters": {},
+        "group": "management_group",
+        "provider": {
+            "name": "systemd",
+            "service": "csm_agent",
+            "intervals": [
+                "0s",
+                "30s",
+                "0s"
+            ],
+            "timeouts": [
+                "60s",
+                "30s",
+                "60s"
+            ],
+            "counters": []
+        },
+        "dependencies": {
+            "predecessors": [
+                "<resource>",
+                "<resource>"
+            ],
+            "colocation": [
+                "<resource>",
+                "<resource>"
+            ],
+            "relation": [
+                "<resource>",
+                "<resource>"
+            ]
+        },
+        "ha": {
+            "mode": "active_passive",
+            "clones": {
+                "active": [
+                    "<max-clone-on-node>",
+                    "<total-clone>"
+                ],
+                "primary": [
+                    "<max-primary-on-node>",
+                    "<total-primary>"
+                ],
+                "parameters": {
+                    "<key1>": "<value1>",
+                    "<key2>": "<value2>"
+                }
+            },
+            "location": {}
+        }
+    },
+    "csm-web": {
+        "parameters": {},
+        "group": "management_group",
+        "provider": {
+            "name": "systemd",
+            "service": "csm_web",
+            "intervals": [
+                "0s",
+                "30s",
+                "0s"
+            ],
+            "timeouts": [
+                "60s",
+                "30s",
+                "60s"
+            ],
+            "counters": []
+        },
+        "dependencies": {
+            "predecessors": [
+                "<resource>",
+                "<resource>"
+            ],
+            "colocation": [
+                "<resource>",
+                "<resource>"
+            ],
+            "relation": [
+                "<resource>",
+                "<resource>"
+            ]
+        },
+        "ha": {
+            "mode": "active_passive",
+            "clones": {
+                "active": [
+                    "<max-clone-on-node>",
+                    "<total-clone>"
+                ],
+                "primary": [
+                    "<max-primary-on-node>",
+                    "<total-primary>"
+                ],
+                "parameters": {
+                    "<key1>": "<value1>",
+                    "<key2>": "<value2>"
+                }
+            },
+            "location": {
+                "<hostname1>": "<priority1>",
+                "<hostname2>": "<priority2>"
+            }
+        }
+    },
+    "kibana": {
+        "parameters": {},
+        "group": "management_group",
+        "provider": {
+            "name": "systemd",
+            "service": "kibana",
+            "intervals": [
+                "0s",
+                "30s",
+                "0s"
+            ],
+            "timeouts": [
+                "60s",
+                "30s",
+                "60s"
+            ],
+            "counters": []
+        },
+        "dependencies": {
+            "predecessors": [
+                "<resource>",
+                "<resource>"
+            ],
+            "colocation": [
+                "<resource>",
+                "<resource>"
+            ],
+            "relation": [
+                "<resource>",
+                "<resource>"
+            ]
+        },
+        "ha": {
+            "mode": "active_passive",
+            "clones": {
+                "active": [
+                    "<max-clone-on-node>",
+                    "<total-clone>"
+                ],
+                "primary": [
+                    "<max-primary-on-node>",
+                    "<total-primary>"
+                ],
+                "parameters": {
+                    "<key1>": "<value1>",
+                    "<key2>": "<value2>"
+                }
+            },
+            "location": {
+                "<hostname1>": "<priority1>",
+                "<hostname2>": "<priority2>"
+            }
+        }
+    }
+}
--- a/ha/core/error.py
+++ b/ha/core/error.py
@@ -36,6 +36,7 @@ HA_CLUSTER_START_ERROR          = 0x000b
 HA_SYSTEM_HEALTH_FAILED         = 0x000c
 HA_CLUSTER_CLI_FAILED           = 0x000d
 HA_EVENT_ANALYZER_ERROR         = 0x000e
+HA_CLUSTER_CONFIG_ERROR         = 0x000f
 
 class HAError(BaseError):
     def __init__(self, rc=1, desc=None, message_id=HA_BASIC_ERROR, message_args=None):
--- a/conf/etc/v2/components/free-space-monitor.json
+++ b/conf/etc/v2/components/free-space-monitor.json
@@ -1,0 +1,57 @@
+{
+    "motr-free-space-mon": {
+        "parameters": { },
+        "group": "monitor_group",
+        "provider": {
+            "name": "systemd",
+            "service": "motr-free-space-monitor",
+            "intervals": [
+                "0s",
+                "30s",
+                "0s"
+            ],
+            "timeouts": [
+                "100s",
+                "30s",
+                "120s"
+            ],
+            "counters": []
+        }, 
+
+        "dependencies": {
+            "predecessors": [
+                "<resource>",
+                "<resource>"
+            ],
+            "colocation": [
+                "<resource>",
+                "<resource>"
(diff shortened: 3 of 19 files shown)


conf/etc/v2/components/csm.json
```diff
@@ -0,0 +1,161 @@
+{
```
indrajitzagade, line 1: @ajaysrivas Not sure we can put these test related files here. Please let us know your thoughts.

azheregelya: @indrajitzagade , @ajaysrivas the thing is that they are not supposed to be test files. Though, I use them in tests.
I tried to fill required information in component json files with some limitations regarding locations and dependencies.

indrajitzagade: ok

indrajitzagade:
Please fix the Codacy issues.

ha/setup/cluster_layout.py
```diff
@@ -0,0 +1,94 @@
...
+Module provides an interface for component files which describe HA cluster.
+"""
+
+import json
+from typing import Dict, List
+from abc import ABC
+from ha.core.error import HAError, HA_CLUSTER_CONFIG_ERROR
+
+
+class ClusterConfigError(HAError):
```
indrajitzagade, line 28: I think lets move this class to error.py file, import and use here.

ha/setup/cluster_status.py
```diff
@@ -0,0 +1,277 @@
...
+import re
+from typing import Callable, List, NamedTuple, Union
+from xml.etree import ElementTree
+from xml.etree.ElementTree import Element
+from cortx.utils.log import Log
+from ha.core.error import HAError, HA_CLUSTER_CONFIG_ERROR
+from ha.execute import SimpleCommand
+
+
+class ClusterStatusError(HAError):
```
indrajitzagade, line 31: I think lets move this class to error.py file, import and use here.

indrajitzagade:
@ajay-paratmandali @shindeam @ArchanaLimaye Please review.

ha/setup/cluster_layout.py
azheregelya: That was the initial idea :)
However I had 2 reservations about it:
1. This validation module is supposed to be provided in separate `.rpm` package. So, it looks logical to ship exceptions together.
2. After reviewing `error.py` I got impression that it became a bucket of boilerplate without real need for that.
When we see `from ha.core.error import ExceptionName` - this doesn't tell a thing about who can issue such exceptions. But when approach `from ModuleName import ExceptionName` is used it becomes clear who is responsible for that exception and in which parts of current module we can expect it. 
So, I took into account 2 these things and decided keep this exception defined inside module. 
As for `error.py` role, I think make sense to keep there only exceptions that are really essential for entire HA, unless there is some requirement I'm not aware of :)

ha/setup/cluster_layout.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    Error during cluster configuration import or validation.
+    """
+    def __init__(self, desc=None):
+        _desc = "HA cluster configuration error" if desc is None else desc
```
knekrasov, line 33: I'm afraid you're assigning to a local variable that will be disposed as soon as `__init__` method is left.

ha/setup/cluster_status.py
```diff
@@ -0,0 +1,276 @@
...
+        assert elem.tag == "resource"
+        b = self._to_bool
+        try:
+            locations = []
+            if int(elem.attrib["nodes_running_on"]) > 0:
+                locations = elem.find("./node").attrib["name"]
+            #  "ocf::heartbeat:Dummy" -> "ocf:heartbeat:Dummy"
+            res_agent = elem.attrib["resource_agent"].split(":")
+            res_agent = "{}:{}:{}".format(res_agent[0], res_agent[2], res_agent[3])
+            result = ClusterResource(
```
knekrasov, line 149: Just return?

ha/setup/cluster_status.py
```diff
@@ -0,0 +1,276 @@
...
+        return result
+
+    def get_resource_from_cloned_group_by_name(self, name: str) -> ClusterResource:
+        """Get resource that is located in cloned group."""
+
+        for group in self.tree.findall("./resources/clone/group"):
+            res = group.find(f"./resource[@id='{name}']")
+            if res is not None:
+                return self._resource_from_xml(res, is_clone=True, group=group.get("id").split(":")[0])
+        return None
```
knekrasov, line 173: Signature doesn't allow None as a return value.

ha/setup/cluster_status.py
```diff
@@ -0,0 +1,276 @@
...
+        res_elems = []
+        for xpath in paths:
+            elems = self.tree.findall(xpath) or []
+            res_elems.extend(elems)
+
+        if len(res_elems) > 1:
+            raise ClusterStatusError(
+                f"Unique resource {name} requested, but several matches found")
+        if not res_elems:
+            return None
```
knekrasov, line 193: None for non-optional return type :-)

ha/setup/cluster_status.py
```diff
@@ -0,0 +1,276 @@
...
+    def get_clone_resource_by_name(self, name: str) -> ClusterCloneResource:
+        """Get clone resource full info."""
+        clone_name = name + "-clone"
+        clone_elem = self.tree.findall(
+            f"./resources/clone[@id='{clone_name}']")
+        if len(clone_elem) > 1:
+            raise ClusterStatusError(
+                f"One clone {name} is expected, but several matches found")
+        if not clone_elem:
+            return None
```
knekrasov, line 259: None for non-optional return type :-)

ha/core/error.py
```diff
@@ -35,6 +35,7 @@
 HA_CLUSTER_START_ERROR          = 0x000b
 HA_SYSTEM_HEALTH_FAILED         = 0x000c
 HA_CLUSTER_CLI_FAILED           = 0x000d
+HA_CLUSTER_CONFIG_ERROR         = 0x000e
```
ajay-paratmandali, line 38: is it unimplemented error ?

ha/setup/validate_cluster.py
```diff
@@ -0,0 +1,24 @@
+import argparse 
+import sys
+from typing import Any
```
ajay-paratmandali, line 3: Do we need separate cli for test or is it only for testing.

If only for integration related test then can we move it in test/integration/setup/

ha/setup/cluster_layout.py
ajay-paratmandali: >This validation module is supposed to be provided in separate .rpm package. So, it looks logical to ship exceptions together.
I think we are not using separate rpm for testing only but it is tested from `ha_setup test`

Can we override it from SetupError instead of HAError
ClusterConfigError It may be use by config manager for confstore as it is cluster config ot ha setup

```
After reviewing error.py I got impression that it became a bucket of boilerplate without real need for that.
When we see from ha.core.error import ExceptionName - this doesn't tell a thing about who can issue such exceptions.
```
I aggrege it has limitation and confusion for use
can we pick this topic in technical discussion and add coding guide line for this

@indrajitzagade  @azheregelya

ha/setup/cluster_layout.py
```diff
@@ -0,0 +1,94 @@
...
+                    layout[resource] = definition
+        return ClusterLayoutJson(nodes=node_list, resources=layout,
+                                 total_num_resources=resource_counter)
+
+    def __init__(self, nodes: List[str], resources: Dict, total_num_resources: int):
+        self.resources = resources
+        self.nodes = nodes
+        self.total_number_of_resources = total_num_resources
+
+    def get_num_nodes(self) -> int:
```
ajay-paratmandali, line 92: We have cluster manager get list of node api can we use that here `get_num_nodes`

ha/setup/cluster_layout.py
azheregelya: > Can we override it from SetupError instead of HAError

Agree. Make sense.

> ClusterConfigError It may be use by config manager for confstore as it is cluster config and ha setup

I can rename it to ValidationConfigError

ha/setup/validate_cluster.py
azheregelya: I would like to clarify this requirement with @ajaysrivas . 
At this moment it looks reasonable to have CLI tools to run some functions on real cluster.

ha/setup/cluster_layout.py
azheregelya: In current case the requirement was to not rely on external code.
I'm currently thinking to even make cluster_validator completely standalone. Even without inheritance from SetupError. Though, i don't see a problem to have cluste_validator -> cortx-ha rpm dependency.

ha/setup/cluster_layout.py
azheregelya: Renamed exceptions and started using SetupError.

ha/setup/cluster_status.py
azheregelya: Added Optional. Thanks.

ha/setup/cluster_status.py
azheregelya: Added Optional. Thanks.

ha/setup/cluster_status.py
azheregelya: Added Optional. Thanks.

ha/setup/cluster_status.py
azheregelya: Fixed. Thanks.

ha/setup/cluster_layout.py
azheregelya: It is passed to `super().__init__ invocation`.

ajaysrivas:
@azheregelya please look for a way to split the PR in multiple PRs.

azheregelya:
PR was split to #382 and #383 PRs

PR #400, opened 2021-05-17
EOS-19624 Password is exposed during add node

akash2144:
Signed-off-by: Akash Dudhane <akash.a.dudhane@seagate.com>

## Problem Statement
<pre>
  <code>
  Story Ref (if any): EOS-19624
    Password is exposed during add node
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  Yes/No
  </code>
</pre>
## Problem Description
<pre>
  <code>
    We are passing command as it is to log file without masking password during add node.
  </code>
</pre>
## Solution
<pre>
  <code>
    Here we are creating copy of command then password is masked in that copied command and passed to log fie.
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
    Unit Testing details here...
  </code>
</pre>

### DIFF
--- a/ha/remote_execution/remote_executor.py
+++ b/ha/remote_execution/remote_executor.py
@@ -30,6 +30,6 @@ class RemoteExecutor:
         self._port = port
 
     @abc.abstractmethod
-    def execute(self, command):
+    def execute(self, command, secret=None):
         '''Communicates to remote node by code execution'''
         raise NotImplementedError
--- a/ha/remote_execution/remote_executor.py
+++ b/ha/remote_execution/remote_executor.py
@@ -30,6 +30,6 @@ class RemoteExecutor:
         self._port = port
 
     @abc.abstractmethod
-    def execute(self, command, secret=None):
+    def execute(self, command: str, secret: str=None):
         '''Communicates to remote node by code execution'''
         raise NotImplementedError
--- a/ha/remote_execution/ssh_communicator.py
+++ b/ha/remote_execution/ssh_communicator.py
@@ -41,7 +41,7 @@ class SSHRemoteExecutor(RemoteExecutor):
             Log.error(f'SSHRemoteExecutor, some error occured while connecting \
                         to SSH channel {err}')
 
-    def execute(self, command: str) -> None:
+    def execute(self, command: str, secret=None) -> None:
         '''
         Communicates to remote node by code execution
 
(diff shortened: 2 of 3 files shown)


ajay-paratmandali (approved):

cortx-admin:
Can one of the admins verify this patch?

ha/remote_execution/ssh_communicator.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, hostname, username=None, port=22, password=None):
             Log.error(f'SSHRemoteExecutor, some error occured while connecting \
                         to SSH channel {err}')
 
-    def execute(self, command: str) -> None:
+    def execute(self, command: str, secret=None) -> None:
```
ajay-paratmandali, line 44: Please check RemoteExecutor parent class of current class and update parent class as 
overridden method should have same definition

akash2144: RemoteExecutor's execute method updated.

ha/remote_execution/remote_executor.py
```diff
@@ -30,6 +30,6 @@ def __init__(self, hostname, port):
         self._port = port
 
     @abc.abstractmethod
-    def execute(self, command):
+    def execute(self, command, secret=None):
```
indrajitzagade, line 33: Why this change?

indrajitzagade:
@akash2144 Please fix the codacy issues.

ha/remote_execution/remote_executor.py
ajay-paratmandali: Akash keep defination same on both class
```
def execute(self, command: str, secret=None) -> None:
def execute(self, command, secret=None):
```
Keep it
def execute(self, command: str, secret: str=None):

akash2144: This is RemoteExecutor parent class's execute method. For overriding, function must have same defination.

akash2144: > Akash keep defination same on both class
> 
> ```
> def execute(self, command: str, secret=None) -> None:
> def execute(self, command, secret=None):
> ```
> 
> Keep it
> def execute(self, command: str, secret: str=None):

Function defination changed to : 
def execute(self, command: str, secret: str=None):

indrajitzagade: Why we need this change @ajay-paratmandali ?

ajaysrivas (approved):

ha/remote_execution/remote_executor.py
ajay-paratmandali: We are overriding `execute` method from parent class 
In overriding method should have same parameter.

If we keep different in parent and child class then we will get error
There is also warning from codacy
```
Parameters differ from overridden 'execute' method
```
as Execute method parameter different in parent and child class

indrajitzagade (approved):

PR #406, opened 2021-05-19
EOS-19565: Cluster and Node start should check and start node if in u…

Madhura-08:
…nstandy state

- After cluster successful start, get the list nodes with the status
- Iterate over this if at least one node is in standby
- Perform unstandby for those nodes

## Problem Statement
<pre>
  <code>
  Story Ref (if any):
    https://jts.seagate.com/browse/EOS-19565
 </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  Yes
  </code>
</pre>
## Problem Description
<pre>
  <code>
   If you stop the cluster while nodes are in standby and if its started again, node remains in standby.   
 </code>
</pre>
## Solution
<pre>
  <code>
    After the cluster start, check what all nodes are in standby and perform standby operation for those 
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
    - Standby the cluster
    - cluster stop using cortx CLI
    - cluster start using cortx CLI
    Cluster was up and running with all nodes in running status
  </code>
</pre>


ha/core/controllers/pcs/cluster_controller.py
```diff
@@ -133,6 +134,25 @@ def start(self) -> dict:
             raise ClusterManagerError(f"Failed to start all nodes {failed_node_list}")
         else:
             status += "All node started successfully, resource start in progress."
+            Log.info('Cluster start is in progress. Now, checking the status of the node. standby or unstandby')
```
ajay-paratmandali, line 137: This should be part of node controller
then cluster_controller is not need to be updated

ajay-paratmandali (changes requested):

ha/core/controllers/pcs/cluster_controller.py
Madhura-08: Sorry. But I feel this is the right place. The following flow explains it better:
1. unstandby the node
2. cluster stop
3. cluster start
After this, first cluster will be started (**pcs cluster start** internally). Then individual node start will be called.
In the node start, first node status will be fetched using **pcs status nodes**. And this is coming as **offline**. 
So, for this, internally **pcs cluster start srvnode-1**  is getting executed. and control will be back to **cluster_executor** start method (from where it started). 
response is : 
 [start] res: {'status': 'InProgress', 'msg': 'Node srvnode-1 : Node was in cluster_offline mode, Cluster started on node successfully'}

So,  I think, its the right position to check this condition. 
Also, instead of performing unstandby for all node one by one, we can perform unstandby using **--all** flag.

Let me know if still there is a point of discussion. 
Thanks !

ajay-paratmandali: Also, instead of performing unstandby for all node one by one, we can perform unstandby using --all flag.

We need to avoide --all flag
it will face problem in 32 node

Madhura-08: Ok.

ha/core/controllers/pcs/cluster_controller.py
```diff
@@ -133,6 +134,22 @@ def start(self) -> dict:
...
+
+            # Get the nodee list with its status
```
indrajitzagade, line 139: *node

ha/core/controllers/pcs/cluster_controller.py
```diff
@@ -133,6 +134,22 @@ def start(self) -> dict:
...
+            node_with_status_dict = self.nodes_status()
+            Log.info(f'########### node status: {node_with_status_dict}')
+
+            # If there is at least one value exists for standby, go ahead and
+            # perform unstandby for all the nodes
+            if NODE_STATUSES.STANDBY.value in node_with_status_dict.values():
+                # Perform the unstandby for all the nodes
+                _output, _err, _rc = self._execute.run_cmd(const.PCS_CLUSTER_UNSTANDBY, check_error=False)
+                if _rc != 0 or _err is not None:
+                    Log.error(f'Cluster started but failed to unstandby the node')
```
indrajitzagade, line 149: I believe we need to send error to the caller if unstandby fails.

indrajitzagade:
@Madhura-08 Please fix codacy issues as well.

ha/core/controllers/pcs/cluster_controller.py
Madhura-08: As per the discussion, moving this code to node_controller so as to avoid one more loop and keep cluster controller separated from this node logic

ha/core/controllers/pcs/cluster_controller.py
Madhura-08: Moved this code to node_controller now.

ha/core/controllers/pcs/cluster_controller.py
Madhura-08: Done. raising an exception there.

ajay-paratmandali (approved):

ajaysrivas:
@ajay-paratmandali 
This code will bring a node to unstandy in cluster start. It will break cluster start - After cluster started, Remove unstandby only if all node in are in standby mode.
Is that correct?

indrajitzagade:
> 
> 
> @ajay-paratmandali
> This code will bring a node to unstandy in cluster start. It will break cluster start - After cluster started, Remove unstandby only if all node in are in standby mode.
> Is that correct?

@ajaysrivas I think above comment is still pending. @ajay-paratmandali

ajaysrivas:
@indrajitzagade It was not implemented earlier too. So behavior is same with respect to existing code. This code looks fine.

indrajitzagade (approved):

PR #417, opened 2021-05-27
EOS-19913: Store node name mapping

indrajitzagade:
## Problem Statement
<pre>
  <code>
  Story Ref (if any):
    EOS-19913: Need to store a node name mapping so that it could be used later for the name conversion
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  Yes
  </code>
</pre>
## Problem Description
<pre>
  <code>
    Need to store a node name mapping so that it could be used later for the name conversion
  </code>
</pre>
## Solution
<pre>
  <code>
    Added code to push the node name mapping to store during mini-provisioning config command.
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
    Tested mini-provisioner config/cleanup command on a single node VM, confirmed that the node name mapping was added/deleted as expected.
  </code>
</pre>


ha/setup/ha_setup.py
```diff
@@ -322,6 +322,11 @@ def process(self):
         self._fetch_fids()
         self._update_cluster_manager_config()
 
+        # Push node name mapping to store
+        if not self._confstore.key_exists(f"{const.PVTFQDN_TO_NODEID_KEY}/{node_name}"):
+            node_id = Conf.get(self._index, f"server_node.{machine_id}.node_id")
+            self._confstore.set(f"{const.PVTFQDN_TO_NODEID_KEY}/{node_name}", node_id)
```
ajay-paratmandali, line 328: const.PVTFQDN_TO_NODEID_KEY}/{node_name}", node_id

I think we need more node map

const.PVTFQDN_TO_NODEID_KEY}/{node_name}" {"hostname": hostname, "node_id":node_id}

can we use json value instead of string

indrajitzagade: @ajay-paratmandali
@ajaysrivas and I had a discussion about this. We do not need to add hostname, instead we will send private_fqdn only in the IEM description as we receive the same from PCS. Hence we need just the private_fqdn to node_id map.

ajay-paratmandali (approved):

ajaysrivas (approved):

PR #441, opened 2021-06-09
EOS-19786: cluster factory reset implementation: Data + Logs

akash2144:
Signed-off-by: Akash Dudhane <akash.a.dudhane@seagate.com>

## Problem Statement
<pre>
  <code>
  Story Ref (if any): EOS-19786
    Cluster factory reset implementation: Data + Logs
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  Yes/No
  </code>
</pre>
## Problem Description
<pre>
  <code>
    Implement code for reset 
    Add management VIP validations.
  </code>
</pre>
## Solution
<pre>
  <code>
    Implemented code for reset.
    Applied logrotate to compress the older version of log file & create new log file. 
    compressed log file are deleted .
    Management VIP validations are already added.
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
    Confirmed that after reset command all logs which are listed in LLD are empty.
  </code>
</pre>


cortx-admin:
Can one of the admins verify this patch?

ha/setup/ha_setup.py
```diff
@@ -632,7 +632,21 @@ def process(self):
...
+        # Delete the logs
+        self.remove_logs(logs)
+        Log.info("reset command is successful")
+
+    def remove_logs(self, logs: list):
+        """
+        Remove logs created by ha.
+        """
+        for log in logs:
+            ResetCmd.remove_file(log)
```
ajaysrivas, line 649: In cleanup, services are still running. If log files are simply deleted then services may not be able to write to log file. I guess, you need to truncate the files and then send SIGHUP signal to syslog server. Please check.

akash2144: Applied logrotate to compress the older version of log file & created new log files. 
compressed log files are deleted.

ha/setup/ha_setup.py
```diff
@@ -633,7 +634,37 @@ def process(self):
...
-        pass # TBD
+        Log.info("Processing reset command")
+        try:
+            # create new version of log file
+            services = ["pcsd", "corosync", "cortx_ha_log.conf", "pacemaker"]
+            for service in services:
+                self._execute.run_cmd(f"logrotate --force /etc/logrotate.d/{service}")
+
+            older_logs = []
+            log_dirs = [const.COROSYNC_LOG, const.RA_LOG_DIR, "/var/log/pcsd"]
```
ajaysrivas, line 645: put /var/log/pcsd in constants like other in the list.

ha/setup/ha_setup.py
```diff
@@ -633,7 +634,37 @@ def process(self):
...
+            for log_dir in log_dirs:
+                log_list = self._execute.run_cmd(f"ls -I *.log -I *.xml {log_dir}")
```
ajaysrivas, line 647: why is command being used here. For listing the file use os module of python and files can be found out from there.

ha/setup/ha_setup.py
```diff
@@ -633,7 +634,37 @@ def process(self):
...
+                logs = log_list[0].splitlines()
+                for log_file in logs:
+                    older_logs.append(os.path.join(log_dir, log_file))
+
+            pacemaker_log_list = self._execute.run_cmd("find /var/log -name pacemaker* -not -name *.log")
```
ajaysrivas, line 652: Again, we should use commands if we don't find a python package for this. Use os module.

ajaysrivas (changes requested):

ha/setup/ha_setup.py
akash2144: Created const as "PCSD_DIR" for /var/log/pcsd

ha/setup/ha_setup.py
akash2144: Used python os module here

ha/setup/ha_setup.py
akash2144: Used python os module here

ajay-paratmandali (approved):

ajaysrivas (approved):

PR #471, opened 2021-06-23
EOS-17922: Add hooks for pre and post disruptive upgrade routines

azheregelya:
## Problem Statement
<pre>
  <code>
  Story Ref (if any):
  EOS-17922
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  No
  </code>
</pre>
## Problem Description
<pre>
  <code>
No pre-disruptive upgrade hooks implemented.
  </code>
</pre>
## Solution
<pre>
  <code>
Provisioner setup.yaml format supports configurable pre and post hooks
for upgrade procedure.
Respective commands were added to HA mini provisioner.
So, perform_post_upgrade was moved from `upgrade` command to new
`post_upgrade` command.

Cluster-related changes in pacemaker are done only when mini-provisioner
is invoked on cluster level (provisioner env variables are being
checked).
Configuration backup is done on every node (node level).
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
This code requires QA testing with upgrade scenario.
Provisioner commands can be used to verify which commands will be invoked for each level.
See: https://github.com/Seagate/cortx-prvsnr/blob/pre-cortx-1.0/docs/mini_prvsnr_api.md#mini-provisioner-commands
  </code>
</pre>

### DIFF
--- a/conf/mini_provisioner/v2/setup.yaml
+++ b/conf/mini_provisioner/v2/setup.yaml
@@ -37,6 +37,14 @@ ha:
     upgrade:
         cmd: /opt/seagate/cortx/ha/bin/ha_setup upgrade
         args: --config $URL
+        pre:
+            cmd: /opt/seagate/cortx/ha/bin/ha_setup pre_upgrade
+            args: --config $URL
+            when: {{ flow == 'upgrade-offline' }}
+        post:
+            cmd: /opt/seagate/cortx/ha/bin/ha_setup post_upgrade
+            args: --config $URL
+            when: {{ flow == 'upgrade-offline' }}
 
     reset:
         cmd: /opt/seagate/cortx/ha/bin/ha_setup reset
--- a/ha/setup/ha_setup.py
+++ b/ha/setup/ha_setup.py
@@ -816,7 +816,6 @@ class PreUpgradeCmd(Cmd):
             if os.environ['PRVSNR_MINI_LEVEL'] == 'cluster':
                 Log.info("Performing pre disruptive upgrade routines on cluster \
                          level")
-                cluster_standby_mode()
                 delete_resources()
             else:
                 Log.info("Performing pre disruptive upgrade routines on node \
--- a/ha/setup/ha_setup.py
+++ b/ha/setup/ha_setup.py
@@ -34,7 +34,7 @@ from ha.const import STATUSES
 from ha.setup.create_pacemaker_resources import create_all_resources
 from ha.setup.post_disruptive_upgrade import perform_post_upgrade
 from ha.setup.pre_disruptive_upgrade import (backup_configuration,
-                                             cluster_standby_mode, delete_resources)
+                                             delete_resources)
 from ha.core.cluster.cluster_manager import CortxClusterManager
 from ha.core.config.config_manager import ConfigManager
 from ha.remote_execution.ssh_communicator import SSHRemoteExecutor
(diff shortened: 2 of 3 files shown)


ajaysrivas:
@azheregelya Please look at following things -
1. Build is broken, probably due to partial check-in.
2. Codacy errors.

ajaysrivas:
@azheregelya Sorry. I didn't check it earlier. This code needs to go into upgrade branch, not in main. Could you please raise PR against upgrade branch.

azheregelya:
> @azheregelya Sorry. I didn't check it earlier. This code needs to go into upgrade branch, not in main. Could you please raise PR against upgrade branch.

will do. thanks for letting know.

azheregelya:
@ajaysrivas changed base branch to `upgrade`

ajaysrivas:
@andkononykhin2 @nitin-seagate Please review. cc: @jaikumargidwani @s-arya

indrajitzagade (approved):

conf/mini_provisioner/v2/setup.yaml
```diff
@@ -37,6 +37,14 @@ ha:
     upgrade:
         cmd: /opt/seagate/cortx/ha/bin/ha_setup upgrade
         args: --config $URL
+        pre:
+            cmd: /opt/seagate/cortx/ha/bin/ha_setup pre_upgrade
+            args: --config $URL
```
andkononykhin2, line 42: currently the same URL as for deployment is passed here but  nothing special is added in scope of the upgrade

ha/setup/ha_setup.py
```diff
@@ -791,6 +790,69 @@ def process(self):
...
+        NOTE: Consul is not expected to be upgraded, so no need to backup
+        consul config
+        Configuration backup is performed for every node while pacemaker setup
+        only on one node (cluster level).
+        """
+        try:
+            if os.environ['PRVSNR_MINI_LEVEL'] == 'cluster':
+                Log.info("Performing pre disruptive upgrade routines on cluster \
+                         level")
+                cluster_standby_mode()
```
andkononykhin2, line 819: cluster level pre-upgrade hook for offline upgrade is called at the time when cluster is already in standby mode. Is it safe to standby again here?

ha/setup/ha_setup.py
```diff
@@ -791,6 +790,69 @@ def process(self):
...
+
+    def process(self):
+        """
+        Process command.
+        """
+        try:
+            if os.environ['PRVSNR_MINI_LEVEL'] == 'cluster':
+                Log.info("Performing post disruptive upgrade routines on cluster \
+                         level")
+                perform_post_upgrade(self.s3_instance)
```
andkononykhin2, line 851: as I see [here](https://github.com/Seagate/cortx-ha/blob/9fc5ba6dc1a3ceb78e84b46ee8daaf9a6afa2511/ha/setup/post_disruptive_upgrade.py#L163) unstandby would be called as part of post-upgrade. Generally it's not accurate since other components may rely on the documented sequence where unstandby/start is called after post-upgrade calls are done for all the components.

ha/setup/ha_setup.py
azheregelya: Technically it is safe, but I can just remove this invocation and rely that Provisioner will put cluster to standby.

ha/setup/ha_setup.py
azheregelya: Updated post upgrade script. Thanks.

conf/mini_provisioner/v2/setup.yaml
azheregelya: I kept it similar to other commands without specific reason. This won't affect mini-provisioner behavior I think.

andkononykhin2 (approved):

ajaysrivas (approved):

PR #480, opened 2021-07-01
EOS-20947: Cortx-HA : mini provisioning cleanup command error if ran …

Madhura-08:
…multiple times.

- Before cleanup starts, check if HA conf file is present or not.
  If not, populate the error message to the use and exit.

## Problem Statement
<pre>
  <code>
  https://github.com/Seagate/cortx-ha/pull/new/EOS-20947
  HA Cleanup fails if performed repeatedly after first successful execution 
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  Yes
  </code>
</pre>
## Problem Description
<pre>
  <code>
  HA Cleanup is failing if performed again after successful execution
  </code>
</pre>
## Solution
<pre>
  <code>
   It is failing because it tries to load HA conf file first which will not be present on the system as it was removed by earlier cleanup. Hence, checking its presence before cleanup starts and if not present, exit with appropriate message is one probable solution
     </code>
</pre>
## Unit Test Cases
<pre>
  <code>
    - Perform cleanup again after first successful execution and check the error message prompted to user.
   </code>
</pre>


ha/setup/ha_setup.py
```diff
@@ -855,6 +855,16 @@ def main(argv: list):
             log_path = Conf.get(const.HA_GLOBAL_INDEX, f"LOG{_DELIM}path")
             log_level = Conf.get(const.HA_GLOBAL_INDEX, f"LOG{_DELIM}level")
             Log.init(service_name='ha_setup', log_path=log_path, level=log_level)
+        elif sys.argv[1] == "cleanup":
+            if not os.path.exists('const.HA_CONFIG_FILE'):
+                a_str = f'Cleanup can not be proceed as \
```
ajay-paratmandali, line 860: Can we change error to warning 

Also is return code 0 or 1

ha/setup/ha_setup.py
```diff
@@ -855,6 +855,16 @@ def main(argv: list):
...
+                           HA config file: {const.HA_CONFIG_FILE} \
+                           is missing. Either cleanp is already done or there \
```
indrajitzagade, line 862: *cleanup

ha/setup/ha_setup.py
```diff
@@ -855,6 +855,16 @@ def main(argv: list):
...
+                           is some other problem'
+                Log.error(f'{a_str}')
```
indrajitzagade, line 864: Will this log work in all cases?

ha/setup/ha_setup.py
```diff
@@ -855,6 +855,16 @@ def main(argv: list):
...
+                sys.stderr.write(a_str)
+                return
+            ConfigManager.init("ha_setup")
```
indrajitzagade, line 867: Why ConfigManager.init() here?

ha/setup/ha_setup.py
Madhura-08: Sure. Will change.

ha/setup/ha_setup.py
Madhura-08: Don't we require init for cleanup where ha_conf file is not missing ?
Its the part when option is cleanup and conf file is present.

ha/setup/ha_setup.py
Madhura-08: Didn't get your question actually.
This will be logged only in cleanup case where cleanup is requested 2nd time (where conf file is missing).

ha/setup/ha_setup.py
Madhura-08: Done.

ha/setup/ha_setup.py
Madhura-08: Done

ha/setup/ha_setup.py
```diff
@@ -855,6 +855,16 @@ def main(argv: list):
...
             Log.init(service_name='ha_setup', log_path=log_path, level=log_level)
+        elif sys.argv[1] == "cleanup":
+            if not os.path.exists('const.HA_CONFIG_FILE'):
+                a_str = f'Cleanup can not be proceed as \
+                           HA config file: {const.HA_CONFIG_FILE} \
+                           is missing. Either cleanup is already done or there \
+                           is some other problem'
+                Log.warn(f'{a_str}')
+                sys.stderr.write(a_str)
+                return 1
```
ajay-paratmandali, line 866:
```
sys.stderr.write(a_str)
```
change to 
```
sys.stdout.write(f"Warning: {a_str})
```
It is warning so return code need to 0

Madhura-08: Done

indrajitzagade (approved):

ha/setup/ha_setup.py
ArchanaLimaye: Not clear on this.... If the option is cleanup and when the ha.conf file is  present, why do we need the init() ?? 
We are nyways going to remove everything right ?
Won't the next "config" CLI when issued, take care of the init() ?

ArchanaLimaye:
Can you please update the
-  "Problem Statement" with the actual problem and 
-  "Unit testing with RPM" sections with updates ; if any

ha/setup/ha_setup.py
Madhura-08: Actually this was already their earlier. For cleanup also, init was executing.
So, ConfigManager init() performs two things:
1. Load HA conf file.
2. Initializes the logs if not already.
Let me check the cleanup without init. How that works and will get back to you on this !

ha/setup/ha_setup.py
indrajitzagade: I mean will the logging work here in this case?

Madhura-08: I got it now. We can't use this Log here. Because it will not be initialized here. Hence removing now.

ha/setup/ha_setup.py
Madhura-08: I am keeping this init part here as cleanup logs will be logged and logging initialization will happen here. If you still have any concerns, let me know. And otherwise also, this part was already present earlier also.

ha/setup/ha_setup.py
Madhura-08: Removing the part to log the warning. Hence, now keeping return code as 1.

indrajitzagade: Instead of the hard-coded '1' here, can we use some standard error code from errno?

ArchanaLimaye (approved):

ha/setup/ha_setup.py
ajay-paratmandali: I think it should be 0

as it is warning not error.
Also if we are returning error then this bug still exists
```
Cleanup command fails if run multiple time
```

Madhura-08: Ok. Returning 0 now

ajay-paratmandali (approved):

ha/setup/ha_setup.py
indrajitzagade: @ajay-paratmandali  Lets resolve this conversation if ok.

PR #487, opened 2021-07-06
Merge cluster status feature into stable branch

indrajitzagade:
Merge cluster status feature into stable branch


ArchanaLimaye:
One generic comment for the bigger merges:
can we  list the files which had merge conflicts that needed to be manually resolved; in the description ? 
So more attention can be provided to those

indrajitzagade:
> 
> 
> One generic comment for the bigger merges:
> can we list the files which had merge conflicts that needed to be manually resolved; in the description ?
> So more attention can be provided to those

Yes, good suggestion. This time there were no merge conflicts to be resolved manually.

ha/core/cluster/cluster_manager.py
```diff
@@ -304,17 +305,16 @@ def controller_list(self) -> list:
...
         Return health status for the requested elements.
-
         Args:
             element ([CLUSTER_ELEMENTS]): The element whose health status is to be returned.
             depth ([int]): A depth of elements starting from the input "element" that the health status
                 is to be returned.
             **kwargs([dict]): Variable number of arguments that are used as filters,
                 e.g. "id" of the input "element".
-
         Returns:
```
ArchanaLimaye, line 318: This return description needs to be changed since we are raising exception instead of a error return value.
I know it is not in the scope of this PR but we can cover it in some upcoming PR ; So whoever is calling this does not need to do additional handling for error values

ajaysrivas (approved):

ha/core/system_health/system_health_exception.py
```diff
@@ -0,0 +1,23 @@
...
+# GNU Affero General Public License for more details.
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <https://www.gnu.org/licenses/>.
+# For any questions about this software or licensing,
+# please email opensource@seagate.com or cortx-questions@seagate.com.
+
+from ha.core.error import HaSystemHealthException
+
+class HealthNotFoundException(HaSystemHealthException):
+    """Exception to indicate health missing for element."""
```
ArchanaLimaye, line 20: Just want to confirm that no code is missed out here in merge (since class is empty )

ArchanaLimaye (approved):

ha/core/system_health/system_health_exception.py
indrajitzagade: yeah, nothing missed. Base class has the code.

ha/core/cluster/cluster_manager.py
indrajitzagade: Still we return error values, not from here but from the called classes.

PR #524, opened 2021-07-26
Changes for EOS-22165 Implement event manager interface to list (get_events) and get message type APIs

ArchanaLimaye:
## Problem Statement
<pre>
  <code>
  Story Ref (if any):
    Need to implement the functions get_events, message_type
  </code>
</pre>
## Unit testing on RPM done
<pre>
  <code>
  No , testing done on deployed setup using python files directly
  </code>
</pre>
## Problem Description
<pre>
  <code>
    Implement the functions get_events, message_type
  </code>
</pre>
## Solution
<pre>
  <code>
    Implemented functionality for get_events, message_type
  </code>
</pre>
## Unit Test Cases
<pre>
  <code>
   On a 3 node deployment  ddded keys to consul directly.
   Tested by executing the event_manager.py file directly that 
  For both function calls
  -  Correct value returned when matching key is found
  -  None / empty list returned when exact matching key is not found
  -  All success / failure paths in code tested ( single key / multiple keys with same prefix / no matching keys present in consul etc)
  </code>
</pre>

### DIFF
--- a/ha/core/event_manager/event_manager.py
+++ b/ha/core/event_manager/event_manager.py
@@ -336,8 +336,8 @@ class EventManager:
         value = []
         Log.debug(f"Fetching subscribed events for {key}")
 
-        if self._confstore.key_exists(key):
-            kv = self._confstore.get(key)
+        kv = self._confstore.get(key)
+        if kv:
             for k, v in kv.items():
                 if k.endswith(key):
                     value = json.loads(v)
@@ -382,8 +382,8 @@ class EventManager:
         value = None
         Log.debug(f"Fetching message type for {key}")
 
-        if self._confstore.key_exists(key):
-            kv = self._confstore.get(key)
+        kv = self._confstore.get(key)
+        if kv:
             for k, v in kv.items():
                 if k.endswith(key):
                     value = v
--- a/ha/core/event_manager/event_manager.py
+++ b/ha/core/event_manager/event_manager.py
@@ -332,9 +332,20 @@ class EventManager:
         Returns:
             list: List of events.
         """
-        pass
+        key = f'{const.COMPONENT_KEY}/{component}'
+        value = []
+        Log.debug(f"Fetching subscribed events for {key}")
 
-    def publish(self, event: RecoveryActionEvent) -> None:
+        if self._confstore.key_exists(key):
+            kv = self._confstore.get(key)
+            for k, v in kv.items():
+                if k.endswith(key):
+                    value = json.loads(v)
+                    break
+        return value
+
+
+    def publish(self, component:str, event: RecoveryActionEvent) -> None:
         """
         Publish event.
         Args:
@@ -367,4 +378,16 @@ class EventManager:
         Args:
             component (str): component name.
         """
-        pass
+        key = const.EVENT_MGR_MESSAGE_TYPE_KEY.replace("<component_id>", component)
+        value = None
+        Log.debug(f"Fetching message type for {key}")
+
+        if self._confstore.key_exists(key):
+            kv = self._confstore.get(key)
+            for k, v in kv.items():
+                if k.endswith(key):
+                    value = v
+                    break
+        return value
+
+# TODO: Add unit tests for this
--- a/ha/core/event_manager/event_manager.py
+++ b/ha/core/event_manager/event_manager.py
@@ -340,7 +340,7 @@ class EventManager:
         if kv:
             for k, v in kv.items():
                 if k.endswith(key):
-                    value = json.loads(v)
+                    value.extend(json.loads(v))
                     break
         return value
 
--- a/ha/util/consul_kv_store.py
+++ b/ha/util/consul_kv_store.py
@@ -144,6 +144,8 @@ class ConsulKvStore:
         self._consul.kv.put(self._prepare_key(key), new_val)
         return new_val
 
+    # TODO : Currently all keys with the matching prefix "key" are returned
+    # add additional parameter so that only the exact matching key is returned when required
     def get(self, key: str = ""):
         """
         Get values. Default it will return all keys. It is block call,


ha/core/event_manager/event_manager.py
```diff
@@ -332,4 +345,16 @@ def message_type(self, component: str) -> str:
         Args:
             component (str): component name.
         """
+        key = const.EVENT_MGR_MESSAGE_TYPE_KEY.replace("<component_id>", component)
+        value = None
+        Log.debug(f"Fetching message type for {key}")
+
```
ajay-paratmandali, line 351: kv = self._confstore.get(key)
if kv is None:
      return kv
else:
      ....

ha/core/event_manager/event_manager.py
```diff
@@ -332,4 +345,16 @@ def message_type(self, component: str) -> str:
...
+        if self._confstore.key_exists(key):
+            kv = self._confstore.get(key)
```
ajay-paratmandali, line 353: for loop not need as key will return only one key val pair 

is single component will have multiple message type

ha/core/event_manager/event_manager.py
```diff
@@ -318,6 +318,19 @@ def get_events(self, component : str) -> list:
...
 
+        key = f'{const.COMPONENT_KEY}/{component}'
+        value = []
+        Log.debug(f"Fetching subscribed events for {key}")
+
+        if self._confstore.key_exists(key):
+            kv = self._confstore.get(key)
+            for k, v in kv.items():
+                if k.endswith(key):
+                    value = json.loads(v)
```
ajay-paratmandali, line 329: value = []
```
value = json.loads(v)
```
assigned again is it append

can we use 
return kv.values()

ajay-paratmandali (changes requested):

ha/core/event_manager/event_manager.py
ArchanaLimaye: There can be multiple keys with same prefix (not expected but keys can get added my mistake) and we want the exact matching key ex   "events/subscribe/sspl"  ,  "events/subscribe/sspl_test" etc
So added the for loop

ha/core/event_manager/event_manager.py
ArchanaLimaye: kv.values returns a string and that needs to be converted

ha/core/event_manager/event_manager.py
ArchanaLimaye: Done

ha/core/event_manager/event_manager.py
```diff
@@ -318,6 +318,19 @@ def get_events(self, component : str) -> list:
             list: List of events.
         """
 
+        key = f'{const.COMPONENT_KEY}/{component}'
```
ajay-paratmandali, line 321: key = f'{const.COMPONENT_KEY}/{component}'

for all key use _prepare_key()
all keys should be in one place to make it easy to maintain

ha/core/event_manager/event_manager.py
```diff
@@ -332,4 +345,16 @@ def message_type(self, component: str) -> str:
         Args:
             component (str): component name.
         """
+        key = const.EVENT_MGR_MESSAGE_TYPE_KEY.replace("<component_id>", component)
```
ajay-paratmandali, line 348:
```
def publish(self, component:str, event: RecoveryActionEvent) -> None:
```
return type not matching
i think publish job is to send event in message bus

ha/core/event_manager/event_manager.py
ajay-paratmandali:
```
def get_events(self, component : str) -> list:
```
and
```
value = json.loads(v)
```
both have different return type

ha/core/event_manager/event_manager.py
```diff
@@ -318,6 +318,19 @@ def get_events(self, component : str) -> list:
             list: List of events.
         """
 
+        key = f'{const.COMPONENT_KEY}/{component}'
+        value = []
+        Log.debug(f"Fetching subscribed events for {key}")
+
+        kv = self._confstore.get(key)
+        if kv:
+            for k, v in kv.items():
```
ajay-paratmandali, line 327: In both function do we need for loop 

are they returning multiple value

ha/core/event_manager/event_manager.py
ArchanaLimaye: this is message_type function , not publish

ha/core/event_manager/event_manager.py
ArchanaLimaye: key creation here is based on the # defines done in earlier PRs.
We can have a generic key creation for all functions , not just get_events and message_type

That can be done in this PR or separately @ajay-paratmandali @indrajitzagade any suggestions  ?

ha/core/event_manager/event_manager.py
ArchanaLimaye: addressing case where keys with same prefix might get added 
ex: "message_type/sspl  , "message_type/sspl_test"  etc
OR the case where passed component string is incomplete ex:  function call looks like message_type("har")
       This should not return the list meant for "hare"

ha/core/event_manager/event_manager.py
ArchanaLimaye: Lets discuss, I am not clear

ha/core/event_manager/event_manager.py
ArchanaLimaye: _confstore.get can return multiple KVs ; where  "key" is the matching prefix. So for loop is needed

ha/core/event_manager/event_manager.py
indrajitzagade: Yes, I agree. we need generic key creation.

ajay-paratmandali (approved):

indrajitzagade (approved):

indrajitzagade:
@ArchanaLimaye  Please resolve conflict.

ArchanaLimaye:
conflicts are resolved

PR #574, opened 2021-10-04
EOS-24820 : fix for ha_client group permissions

shindeam:
Signed-off-by: Amol Shinde <amol.shinde@seagate.com>

# Problem Statement
- EOD-24820 User created with haclient group throws error

# Design
-  For Bug, Describe the fix here.
-  For Feature, Post the link for design

# Coding
-  [ ] Coding conventions are followed and code is consistent

# Testing 
- [ ] Testing was performed with RPM

# Review Checklist 
- [ ] PR is self reviewed
- [ ] JIRA number/GitHub Issue added to PR
- [ ] Jira and state/status is updated and JIRA is updated with PR link

# Review Checklist 

# Documentation
  Checklist for Author

### DIFF
--- a/ha/const.py
+++ b/ha/const.py
@@ -23,6 +23,10 @@ HA_CLUSTER_SOFTWARE="corosync"
 HACLUSTER_KEY = "cortx"
 SERVER_NODE_KEY = "server_node"
 RA_LOG_DIR="/var/log/seagate/cortx/ha"
+RA_LOG_FILES = [f"{RA_LOG_DIR}/alert_monitor.log", f"{RA_LOG_DIR}/cluster_manager.log",
+                f"{RA_LOG_DIR}/event_analyzerd.log", f"{RA_LOG_DIR}/event_manager.log",
+                f"{RA_LOG_DIR}/health_monitor.log"]
+
 PACEMAKER_LOG="/var/log/pacemaker.log"
 AUTH_DIR="/var/lib/pcsd"
 PCSD_LOG="/var/log/pcsd/pcsd.log"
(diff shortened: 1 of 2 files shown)


ha/const.py
```diff
@@ -23,6 +23,10 @@
 HACLUSTER_KEY = "cortx"
 SERVER_NODE_KEY = "server_node"
 RA_LOG_DIR="/var/log/seagate/cortx/ha"
+RA_LOG_FILES = [f"{RA_LOG_DIR}/alert_monitor.log", f"{RA_LOG_DIR}/cluster_manager.log",
+                f"{RA_LOG_DIR}/event_analyzerd.log", f"{RA_LOG_DIR}/event_manager.log",
+                f"{RA_LOG_DIR}/health_monitor.log"]
+
```
ArchanaLimaye, line 29: This will not scale in case more log files get added

ha/setup/ha_setup.py
```diff
@@ -401,6 +401,12 @@ def process(self):
             self._execute.run_cmd(f"setfacl -R -m g:{const.USER_GROUP_ROOT}:rwx {const.RA_LOG_DIR}")
             self._execute.run_cmd(f"setfacl -R -m g:{const.USER_GROUP_HACLIENT}:r-x {const.CONFIG_DIR}")
             self._execute.run_cmd(f"setfacl -R -m g:{const.USER_GROUP_ROOT}:rwx {const.CONFIG_DIR}")
+            for file_path in const.RA_LOG_FILES:
+                if file_path and not os.path.exists(file_path):
+                    os.makedirs(os.path.dirname(file_path), exist_ok=True)
+                    open(file_path, 'x')
+                self._execute.run_cmd(f"setfacl -R -m g:{const.USER_GROUP_HACLIENT}:rwx {file_path}")
```
ArchanaLimaye, line 408: This can be done in ConfigManager.init
That way new log files need not be created here.
Also scaling in case of new logs will not be an issue.

ha/const.py
shindeam: I think we can add modules log file names here in the list

ha/setup/ha_setup.py
shindeam: I tried adding it in the ConfigManager.init, But when we try to instantiate SimpleCommand class it fails at Log.debug statement with none value exception.

PR #595, opened 2021-12-06
Eos 26559 [K8S:RAW] Handle comments from system health changes ticket

mukhtar-inamdar:
# Problem Statement
- [K8S:RAW] Handle comments from system health changes ticket
- 2 comments 1 is adding log before starting the poll function
- and another 1 is to check the return value which is returning from the call function of fault tolerance for receiving events from message bus
# Design
-  Once the event received from message bus callback function process it and if failed it returns fails which cause
- to call the same function again with no change but if the failure persists then this call will do an endless loop.
- to avoid this added loop inside system_health.process_event() function will retry for some limited time with wait 50 miliseconds.
- To test added force exception and checked: The function system_health.process_event() retrying 3 times then re-throwing exception which then ignored by fault_tolerance event callback function.
- Also tested positive scenarios as it processing further as it is no failure due to new chnages

# Coding
-  [ ] Coding conventions are followed and code is consistent

# Testing 
- [ ] Testing was performed with RPM

# Review Checklist 
- [ ] PR is self reviewed
- [ ] JIRA number/GitHub Issue added to PR
- [ ] Jira and state/status is updated and JIRA is updated with PR link

# Review Checklist 

# Documentation
  Checklist for Author


mukhtar-inamdar:
wrong pull request created hence deleting

PR #596, opened 2021-12-06
 EOS-26644 [K8S:RAW] Handle comments from system health changes ticket

mukhtar-inamdar:
# Problem Statement
- [K8S:RAW] Handle comments from system health changes ticket
- 2 comments 1 is adding log before starting the poll function
- and another 1 is to check the return value which is returning from the call function of fault tolerance for receiving events from message bus
# Design
-  Once the event received from message bus callback function process it and if failed it returns fails which cause
- to call the same function again with no change but if the failure persists then this call will do an endless loop.
- to avoid this added loop inside system_health.process_event() function will retry for some limited time with wait 50 miliseconds.
- To test added force exception and checked: The function system_health.process_event() retrying 3 times then re-throwing exception which then ignored by fault_tolerance event callback function.
- Also tested positive scenarios as it processing further as it is no failure due to new changes.
- Testing: [EOS-26644_testing.txt](https://github.com/Seagate/cortx-ha/files/7660812/EOS-26644_testing.txt)

# Coding

# Test scenarios
- raised ConsulException from ConfigManager.get_confstore()and the result it returns CONSUMER_STATUS.SUCCESS
- raised ConfError from SystemHealth and the result it returns CONSUMER_STATUS.SUCCESS
- raised EventFilterException from ClusterResourceFilter.filter_event() and the result it returns CONSUMER_STATUS.SUCCESS
- raised EventParserException from ClusterResourceParser.parse_event and the result it returns CONSUMER_STATUS.SUCCESS
- raised SubscriberException from SystemHealth.process_event() and the result it returns CONSUMER_STATUS.FAILED and hence it went into the loop.
- raised unknown exception from ConfigManager.get_confstore()and the result it returns CONSUMER_STATUS.SUCCESS

# Testing 

# Review Checklist 
- [x] PR is self reviewed

# Review Checklist 

# Documentation
  Checklist for Author

### DIFF
--- a/ha/core/event_analyzer/event_analyzerd.py
+++ b/ha/core/event_analyzer/event_analyzerd.py
@@ -37,6 +37,7 @@ from ha.core.system_health.system_health import SystemHealth
 from ha.core.event_analyzer.watcher.watcher import Watcher
 from ha.core.event_analyzer.filter.filter import ClusterResourceFilter
 from ha.core.event_analyzer.parser.parser import ClusterResourceParser
+from ha.core.event_analyzer.event_analyzer_exceptions import SubscriberException
 from ha.const import _DELIM
 
 class EventAnalyzerService:
@@ -113,7 +114,11 @@ class EventAnalyzer:
         self._cluster_resource_parser = ClusterResourceParser()
         if self._cluster_resource_filter.filter_event(msg):
             health_event = self._cluster_resource_parser.parse_event(msg)
-            system_health.process_event(health_event)
+            try:
+                system_health.process_event(health_event)
+            except Exception as e:
+                Log.error(f"Failed to publish event. Error: {e}")
+                raise SubscriberException(f"Failed to process event {str(health_event)}. Error: {e}")
 
 def main(argv):
     """
--- a/ha/core/event_analyzer/event_analyzerd.py
+++ b/ha/core/event_analyzer/event_analyzerd.py
@@ -117,7 +117,7 @@ class EventAnalyzer:
             try:
                 system_health.process_event(health_event)
             except Exception as e:
-                Log.error(f"Failed to publish event. Error: {e}")
+                Log.error(f"Failed to process event. Error: {e}")
                 raise SubscriberException(f"Failed to process event {str(health_event)}. Error: {e}")
 
 def main(argv):
--- a/ha/core/system_health/system_health.py
+++ b/ha/core/system_health/system_health.py
@@ -25,7 +25,7 @@ from cortx.utils.log import Log
 from ha import const
 from ha.util.message_bus import MessageBus
 from cortx.utils.conf_store.conf_store import Conf
-from ha.const import _DELIM
+from ha.const import _DELIM, HA_MAX_RETRY
 from ha.core.config.config_manager import ConfigManager
 from ha.core.system_health.health_evaluators.element_health_evaluator import ElementHealthEvaluator
 from ha.core.system_health.const import EVENT_SEVERITIES, NODE_MAP_ATTRIBUTES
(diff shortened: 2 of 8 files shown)


mukhtar-inamdar:
[EOS-26644_testing.txt](https://github.com/Seagate/cortx-ha/files/7660801/EOS-26644_testing.txt)

shaileshsvc (approved):
Approved. Retry (with timeout and retry count) only if processing errors and need to retry processing with possibility of success.

ha/core/system_health/system_health.py
```diff
@@ -337,60 +337,66 @@ def process_event(self, healthevent: HealthEvent):
         """
         Process Event method. This method could be called for updating the health status.
         """
+        for retry in range(HA_MAX_RETRY):
```
ArchanaLimaye, line 340: Blanket retry without knowing the cause of the exception can cause issues.
ex
If we have updated health of some components already and then we retry the whole sequence, extra updates might happens. This can cause unwanted /extra events to be sent to Hare.
If  health of any module gets updated multiple times un-necessarily, the timestamps will get modified for the same and that can lead to issues.
More such problems might also come. 

Please refer to the code from main branch in file ./ha/core/event_analyzer/watcher/watcher.py
Here the exceptions are getting handled individually. We need to do something similar ether in fault-tolerance_driver.py or in EventAnalyzer() [where we are getting those same exceptions.]

ArchanaLimaye (changes requested):
Please refer to the comment

ha/core/system_health/system_health.py
mukhtar-inamdar: changed as discussed and tested for every exception and found expected behaviour

shaileshsvc:
Include testing details or mention where the testing details are (say in Jira)

shaileshsvc (approved):
Approved.

ArchanaLimaye:
please update the testing details in PR

ArchanaLimaye:
Rebase the branch to latest fault_k8s

ha/core/event_analyzer/event_analyzerd.py
```diff
@@ -113,7 +114,11 @@ def __init__(self, msg=None):
         self._cluster_resource_parser = ClusterResourceParser()
         if self._cluster_resource_filter.filter_event(msg):
             health_event = self._cluster_resource_parser.parse_event(msg)
-            system_health.process_event(health_event)
+            try:
+                system_health.process_event(health_event)
+            except Exception as e:
+                Log.error(f"Failed to publish event. Error: {e}")
```
ArchanaLimaye, line 120: Log should be Failed to process event.

ha/fault_tolerance/fault_tolerance_driver.py
```diff
@@ -61,14 +67,31 @@ def process_message(self, message: str):
...
+            return CONSUMER_STATUS.SUCCESS
+        except EventParserException as e:
+            Log.error(f"Parser exception {e} {traceback.format_exc()} for {message}.  Ack Message.")
+            return CONSUMER_STATUS.SUCCESS
+        except SubscriberException as e:
+            Log.error(f"Subscriber exception {e} {traceback.format_exc()} for {message}, retry without ack.")
+            return CONSUMER_STATUS.FAILED
+        except Exception as e:
+            Log.error(f"Unknown Exception caught {e} {traceback.format_exc()}")
+            Log.error(f"Forcefully ack failed msg: {message}")
```
ArchanaLimaye, line 88: @shaileshsvc should we return SUCCESS here?
is it  better not to retry this message since we don't know what has caused this failure ? This will also be in sync with earlier code at line 67 that was deleted

ha/core/event_analyzer/event_analyzerd.py
mukhtar-inamdar: done

ha/fault_tolerance/fault_tolerance_driver.py
mukhtar-inamdar: As discussed returning success from here.

ArchanaLimaye (approved):

PR #610, opened 2022-01-03
Eos 27021 : sigterm handling

mukhtar-inamdar:
# Problem Statement
- Implement sigterm handling

# Design
-  implement sigterm handling for k8s monitor, fault tolerance monitor, and health monitor
- once sigterm received k8s monitor will flush the remaining k8s events and release quired resources and other monitors will flush pending messages in the message bus.
- after this, all monitors will close their monitor threads.
- where required used timeout value and it is defined as constant '5' and it is same as cluster constant BASE_WAIT_TIME
- added unitest for testing the sigterm hadling
- Separate the join and stop calls of message_bus.py calls messageBusConsumer to make stops every consumer in the same process parallelly.

# Coding

# Test cases and result
[EOS-27021-unittesting.txt](https://github.com/Seagate/cortx-ha/files/7807031/EOS-27021-unittesting.txt)

# Testing 
- [x] Test Cases cover Happy Path, Non-Happy Path, and Scalability

# Review Checklist 
- [X] PR is self reviewed
- [X] JIRA number/GitHub Issue added to PR
- [X] Check if the description is clear and explained
- [] Is there a change in filename/package/module or signature? [Y/N]:  **N**
- [ ] If yes for above point, is a notification sent to all other cortx components? [Y/N] **N**
- [ ] Side effects on other features (deployment/upgrade)? [Y/N] **N**
- [ ] Dependencies on other component(s)? [Y/N] **N**

# Review Checklist 

# Documentation
  Checklist for Author
- [x] Changes done to WIKI / Confluence page / Quick Start Guide
- Yes, will check and update on required places.


shaileshsvc (approved):
Approved, (fix, codacy, update the unit test results if changed).

shaileshsvc (changes requested):
Fix codacy
> @mukhtar-inamdar Done

shaileshsvc (approved):
Approved

Madhura-08 (approved):
Approving. But please add comments wherever applicable and please add function docstring.

ha/core/health_monitor/health_monitord.py
```diff
@@ -110,14 +113,26 @@ def process_event(self, message: str) -> None:
...
+
+    def start(self):
         """
-        Run server
+        Starts consumer daemon thread to receive the alters and perform action on it.
         """
-        Log.info(f"Running the daemon for Health Monitor with PID {os.getpid()}...")
+        Log.info("Starting the daemon for Health Monitor...")
         self._event_consumer.start()
-        while True:
```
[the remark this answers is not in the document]
Madhura-08: Join() will block this thread.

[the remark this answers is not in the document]
akash2144: Ohk. @mukhtar-inamdar  please resolve this comment.

akash2144 (approved):

ha/fault_tolerance/fault_tolerance_driver.py
```diff
@@ -33,33 +31,45 @@ class FaultTolerance:
...
-        wait method for receiving events
+        join and wait for monitor threads to exit
         """
-        Log.debug("FaultTolerance poll")
-        try:
-            while True:
-                time.sleep(self._wait_time)
-        except Exception as exe:
-            raise(f'Oops, some issue in the fault tolerance_driver: {exe}')
+        self.node_fault_monitor.join()
```
ArchanaLimaye, line 67: What happens to threads if there is an exception ? 
This is a generic query for the modified functionality.
Is the behavior same as before in case Exceptions are received in a thread ?

ha/monitor/k8s/monitor.py
```diff
@@ -15,35 +15,104 @@
...
+        # Creating POD monitor object
+        self.pod_monitor = ObjectMonitor(producer, K8SClientConst.POD, **kwargs)
+
+    def _get_producer(self):
+        """
+        Get message bus producer
+        """
+        message_type = Conf.get(const.HA_GLOBAL_INDEX, f"MONITOR{_DELIM}message_type")
+        producer_id = Conf.get(const.HA_GLOBAL_INDEX, f"MONITOR{_DELIM}producer_id")
+        Log.info("Initializing message bus.")
```
ArchanaLimaye, line 81: The Guidance from Ujjwal is that there should be no logs in the system which do not print any variable data. 
Please remove this log, the log on line 83 will suffice; you can append this message to that log if needed. 

Please remove/modify any similar logs from other places also

ha/monitor/k8s/monitor.py
```diff
@@ -15,35 +15,104 @@
...
+        Log.info(f"Getting producer {producer_id} for message type: {message_type}")
+        return MessageBus.get_producer(producer_id, message_type)
+
+    def set_sigterm(self, signum, frame):
+        """
+        Callback function to receive a signal
+        """
+        Log.info(f"Received SIGTERM {signum}.")
+        Log.debug(f"Received signal: {signum} during execution of frame: {frame}")
+        Log.info("Stopping the K8s Monitor...")
```
ArchanaLimaye, line 92: Please remove this log or merge into the previous one

ha/monitor/k8s/monitor.py
```diff
@@ -15,35 +15,104 @@
...
-    kwargs['label_selector'] = f'name in ({pod_label_str})'
-    pod_thread = ObjectMonitor('pod', **kwargs)
-
-    pod_thread.start()
-    node_thread.start()
-
-    object_threads.append(pod_thread)
-    object_threads.append(node_thread)
-
-    for a_thread in object_threads:
```
ArchanaLimaye, line 46: minor comment. 
See if we can have a list of object threads like here.
That way when monitor gets extended to have watchers for containers , PVCs , it will be simpler to itherate over list and join all threads ,

ha/monitor/k8s/object_monitor.py
```diff
@@ -64,20 +98,40 @@ def run(self):
...
+                    else:
+                        self._starting_up = False
+
+                # Write to message bus
+                Log.info(f"{self._object}_monitor Sending alert on message bus {alert.to_dict()}")
+                self._producer.publish(str(alert.to_dict()))
+
+            # If stop processing events is set then no need to retry just break the loop and
+            # If we don't specify timeout no need to restart the loop it will happen internally
+            if self._stop_event_processing or K8SClientConst.TIMEOUT_SECONDS not in self._args:
```
ArchanaLimaye, line 135: 1. When would TIMEOUT_SECONDS not be present in _args  since we are always passing it ?
2. How will loop restart in this condition ?

ha/fault_tolerance/fault_tolerance_driver.py
mukhtar-inamdar: main thread which is joining will wait till all thread joins.
no matter who gets the exception and stops.
the thread will by default print stack-trace but the mail thread which is waiting will not get notified. 

Well, catch! this needs handle but look like the previous while loop was also not getting notify event if it was trying to check exception.

Needs to handle this

ha/util/message_bus.py
```diff
@@ -131,19 +139,48 @@ def run(self):
             except Exception as e:
                 Log.error(f"Supressing exception from message bus {e}")
                 retry = False
+        if self.flush_on_exit:
```
ArchanaLimaye, line 142: Can you please add a comment here saying that we do not expect any messages to be present in the message bus at this point ( since previously received cluster stop would have ensured that message bus is empty) and this is just a precautionary code.

ha/monitor/k8s/monitor.py
mukhtar-inamdar: Done

ha/monitor/k8s/monitor.py
mukhtar-inamdar: done

ha/monitor/k8s/object_monitor.py
mukhtar-inamdar: if tomorrow we don't need to pass timeout in some cases for some specific monitor thread.
in that case, while True is not required as the loop inside the Watch.stream will make sure to continue which will not be the case if we provide the timeout.
so in that case, if we don't provide timeout then no need to retry just break the loop as Watch.stream will be blocking the call which has while True loop so no need for a loop for it.

ha/util/message_bus.py
mukhtar-inamdar: done

ha/fault_tolerance/fault_tolerance_driver.py
mukhtar-inamdar: new issue regarding this: https://jts.seagate.com/browse/EOS-27291

ArchanaLimaye (approved):

PR #645, opened 2022-02-21
CORTX-28560 & CORTX-29057 : Fix duplicate events published from k8s_monitor & Event parser issue

mariyappanp:
# Problem Statement
CORTX-28560 - k8s_monitor is repeating previously sent alert event there is no status change on monitoring objects.
CORTX-29057 - Unpacking issue with Event parser when it returns None

# Design
CORTX-28560 - Fix is to publish only new alert by checking whether incoming alert is already sent or not.
CORTX-29057 - Return required number of values as client code expects

# Coding

# Testing 
    Test is added. Logs are also validated.
- [x] Test Cases cover Happy Path, Non-Happy Path and Scalability
**Log:** https://jts.seagate.com/secure/attachment/505658/DEV_fix_log_23FEB2022.txt

# Review Checklist 
- [x] PR is self reviewed
- [x] Is there a change in filename/package/module or signature? [Y/N]: N
- [x] If yes for above point, is a notification sent to all other cortx components? [Y/N] N.A
- [x] Side effects on other features (deployment/upgrade)? [Y/N] N
- [x] Dependencies on other component(s)? [Y/N] N

# Review Checklist 

# Documentation
  Checklist for Author

### DIFF
--- a/ha/monitor/k8s/object_monitor.py
+++ b/ha/monitor/k8s/object_monitor.py
@@ -182,16 +182,15 @@ class ObjectMonitor(threading.Thread):
         """
         Check incoming alert is already published or not.
         If incoming alert is not found in published alerts, then
-        it is a new alert to publish. Alert will be stored like,
-        self._published_alerts = {
-            resource_name: { generation_id : alert }
-            }
-
+        it is a new alert to publish.
+        Alert will be stored and mapped to its unique key,
+            self._published_alerts = { alert_key : alert }
         Returns:
             True if it is published already
             False if it is a new alert
         """
         incoming_alert = alert.to_dict().copy()
+        incoming_resource_type = incoming_alert.get('_resource_type')
         incoming_resource_name = incoming_alert.get('_resource_name')
         incoming_generation_id = incoming_alert.get('_generation_id')
 
@@ -204,21 +203,23 @@ class ObjectMonitor(threading.Thread):
         if "_is_status" in incoming_alert.keys():
             del incoming_alert['_is_status']
 
-        incoming_alert_msg = json.dumps(incoming_alert, sort_keys=True)
-        published_alerts = self._published_alerts.get(incoming_resource_name)
+        if incoming_generation_id:
+            alert_key = f"{incoming_resource_type}_{incoming_generation_id}"
+        else:
+            alert_key = f"{incoming_resource_type}_{incoming_resource_name}"
 
-        if published_alerts:
-            incoming_generation_id = incoming_alert.get('_generation_id')
-            published_alert = published_alerts.get(incoming_generation_id)
+        incoming_alert_msg = json.dumps(incoming_alert, sort_keys=True)
+        published_alert = self._published_alerts.get(alert_key)
+
+        if published_alert:
             if incoming_alert_msg == published_alert:
                 # Published already
                 return True
             else:
                 # New alert
-                self._published_alerts[incoming_resource_name][incoming_generation_id] = incoming_alert_msg
+                self._published_alerts[alert_key] = incoming_alert_msg
+
         else:
-            self._published_alerts[incoming_resource_name] = {
-                incoming_generation_id: incoming_alert_msg
-                }
+            self._published_alerts[alert_key] = incoming_alert_msg
 
         return False
--- a/ha/monitor/k8s/object_monitor.py
+++ b/ha/monitor/k8s/object_monitor.py
@@ -186,10 +186,10 @@ class ObjectMonitor(threading.Thread):
         self._published_alerts = { resource_name: [alert1, alert2] }
 
         Returns:
-            True if incoming alert is a new alert
-            False, otherwise.
+            True if it is published already
+            False if it is a new alert
         """
-        incoming_alert = alert.to_dict()
+        incoming_alert = alert.to_dict().copy()
 
         # Alert with no status change also has new timestamp.
         # So timestamp field should be ignored for validation.
--- a/ha/monitor/k8s/object_monitor.py
+++ b/ha/monitor/k8s/object_monitor.py
@@ -195,16 +195,13 @@ class ObjectMonitor(threading.Thread):
         incoming_resource_name = incoming_alert.get('_resource_name')
         incoming_generation_id = incoming_alert.get('_generation_id')
 
-        if incoming_alert.get('_timestamp') is None or incoming_resource_name is None:
-            # Consider any other alerts to be published
-            return False
-
         # Alert which is getting repeated also has new timestamp.
         # So timestamp field should be ignored for validation.
-        del incoming_alert['_timestamp']
+        if "_timestamp" in incoming_alert.keys():
+            del incoming_alert['_timestamp']
 
         # Remove user added fields those aren't exist in raw event
-        if incoming_alert.get('_is_status') is not None:
+        if "_is_status" in incoming_alert.keys():
             del incoming_alert['_is_status']
 
         incoming_alert_msg = json.dumps(incoming_alert, sort_keys=True)
--- a/ha/monitor/k8s/object_monitor.py
+++ b/ha/monitor/k8s/object_monitor.py
@@ -199,7 +199,7 @@ class ObjectMonitor(threading.Thread):
             # Consider any other alerts to be published
             return False
 
-        # Alert with no status change also has new timestamp.
+        # Alert which is getting repeated also has new timestamp.
         # So timestamp field should be ignored for validation.
         del incoming_alert['_timestamp']
 
--- a/ha/monitor/k8s/object_monitor.py
+++ b/ha/monitor/k8s/object_monitor.py
@@ -15,6 +15,7 @@
 # about this software or licensing, please email opensource@seagate.com or
 # cortx-questions@seagate.com.
 
+import json
 import threading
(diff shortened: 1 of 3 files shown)


ArchanaLimaye:
Please add testing details

ha/monitor/k8s/object_monitor.py
```diff
@@ -167,3 +172,44 @@ def run(self):
...
+        resource_name = incoming_alert.get('_resource_name')
+
+        if resource_name:
+            incoming_alert_msg = json.dumps(incoming_alert, sort_keys=True)
+            published_alerts = self._published_alerts.get(resource_name)
+            if published_alerts and isinstance(published_alerts, list):
+                if incoming_alert_msg in published_alerts:
+                    return True
+                else:
+                    published_alerts.append(incoming_alert_msg)
```
ArchanaLimaye, line 207: We need to store only one alert per resource_name right ?  Why do we need a list ?

ha/monitor/k8s/object_monitor.py
```diff
@@ -167,3 +172,44 @@ def run(self):
...
+
+        Returns:
+            True if it is published already
+            False if it is a new alert
+        """
+        incoming_alert = alert.to_dict().copy()
+
+        # Alert with no status change also has new timestamp.
+        # So timestamp field should be ignored for validation.
+        if not incoming_alert.get('_timestamp') is not None:
```
ArchanaLimaye, line 190: Can you clarify this check ?

ha/monitor/k8s/object_monitor.py
mariyappanp: During pod deletion, current pod is getting terminated and new pod is getting created.
So there are two pods events one by one per resource_name. That's why list is used.

If I don't use list,
1. New pod creation event will be caught initially (published - online)
2. Pod which is getting deleted will overwrite that (published -failed)
This would still send that new pod (online) event which will become duplicate of (1).

ha/monitor/k8s/object_monitor.py
mariyappanp: Even there is no change in status change, the watch stream repeats recent status in event at every 5 seconds with new timestamp. This is expected from watch stream. Rather than saying its an event, better can say its a response (as variable name defined for that in kubernetes watch stream module).

When no new event, but response is coming from stream, published and this new event is same and only the _timestamp differ. In this case I'm ignoring this _timestamp filed before the alert messages comparison.

ha/monitor/k8s/object_monitor.py
mariyappanp: There is an another scenario I need to cover. I will fulfil the gap in next commit.

mariyappanp: @ArchanaLimaye can you check now.

ha/monitor/k8s/object_monitor.py
mariyappanp: @ArchanaLimaye , correcting this comment as below may give more clarity. Fine?
```
        # Repeating alert also has new timestamp on same event.
        # So timestamp field should be ignored for validation.
```

mariyappanp:
> Please add testing details
Logs are verified. Log file is attached to the bug. 
https://jts.seagate.com/secure/attachment/505658/DEV_fix_log_23FEB2022.txt

ha/monitor/k8s/object_monitor.py
```diff
@@ -167,3 +172,50 @@ def run(self):
                 break
         Log.info(f"Stopping the {self.name}...")
 
+    def _is_published_alert(self, alert) -> bool:
+        """
+        Check incoming alert is already published or not.
+        If incoming alert is not found in published alerts, then
+        it is a new alert to publish. Alert will be stored like,
```
indrajitzagade, line 179: How many such alerts will be stored?

mariyappanp: It stores one alert per pod.
But there can be more alerts stored depends on how many pod created and deleted during k8s_monitor service run time.

ha/monitor/k8s/object_monitor.py
```diff
@@ -167,3 +172,50 @@ def run(self):
...
+        self._published_alerts = {
+            resource_name: { generation_id : alert }
+            }
+
+        Returns:
+            True if it is published already
+            False if it is a new alert
+        """
+        incoming_alert = alert.to_dict().copy()
```
Madhura-08, line 188: Please rebase. alert is not getting used now. You can use Event class which is in fault tolerance right now. Will be moved to utils

ha/monitor/k8s/object_monitor.py
```diff
@@ -167,3 +172,50 @@ def run(self):
...
+        incoming_resource_name = incoming_alert.get('_resource_name')
+        incoming_generation_id = incoming_alert.get('_generation_id')
```
Madhura-08, line 190: Also, now the alert structure is also changed. So, this get call will also change

ha/monitor/k8s/object_monitor.py
```diff
@@ -167,3 +172,50 @@ def run(self):
                 break
         Log.info(f"Stopping the {self.name}...")
 
+    def _is_published_alert(self, alert) -> bool:
+        """
+        Check incoming alert is already published or not.
+        If incoming alert is not found in published alerts, then
+        it is a new alert to publish. Alert will be stored like,
+        self._published_alerts = {
+            resource_name: { generation_id : alert }
```
Madhura-08, line 181: For disk alert, generation id will not be there (will be something else may be). So, how disk alerts will be stored?

ha/monitor/k8s/object_monitor.py
mariyappanp: Done.

ha/monitor/k8s/object_monitor.py
mariyappanp: Done.

ha/monitor/k8s/object_monitor.py
```diff
@@ -167,3 +172,47 @@ def run(self):
...
+
+        # Remove user added fields those aren't exist in raw event
+        if "_is_status" in incoming_alert.keys():
+            del incoming_alert['_is_status']
+
+        incoming_alert_msg = json.dumps(incoming_alert, sort_keys=True)
+        published_alerts = self._published_alerts.get(incoming_resource_name)
+
+        if published_alerts:
+            incoming_generation_id = incoming_alert.get('_generation_id')
```
Madhura-08, line 205: I think you have this value already. Line no 190.

ha/monitor/k8s/object_monitor.py
mariyappanp: Can you provide the reference?

ha/monitor/k8s/object_monitor.py
mariyappanp: With the new commit, 
`self._published_alerts = { alert_key : alert }`
I handled this case,
```
        if incoming_generation_id:
            alert_key = f"{incoming_resource_type}_{incoming_generation_id}"
        else:
            alert_key = f"{incoming_resource_type}_{incoming_resource_name}"
```
Host alert also doesn't have generation id. But resource name is unique there.
In case Disk alert won't have any unique value, then we need to try other way.
At this time this is sufficient.

mukhtar-inamdar (commented):
please check Madhuri's comments,
otherwise, the logic looks good.
Q. What if for a particular pod if offline and online both events repeated.
have you checked to restart the same pod multiple times during a few minutes gap of you can test the same with mock also?

ha/monitor/k8s/object_monitor.py
mariyappanp: Done.

ha/monitor/k8s/object_monitor.py
mariyappanp: Done.

ha/monitor/k8s/object_monitor.py
mariyappanp: `Note: Disk monitoring is not covered by k8s monitor`
However similar case suits for host alert.

Done.

mariyappanp:
> please check Madhuri's comments, otherwise, the logic looks good. Q. What if for a particular pod if offline and online both events repeated. have you checked to restart the same pod multiple times during a few minutes gap of you can test the same with mock also?

Yes, acknowledged.

ArchanaLimaye (approved):

mukhtar-inamdar (approved):

ha/monitor/k8s/object_monitor.py
Madhura-08: OK..Yes..Got it.

ha/monitor/k8s/object_monitor.py
Madhura-08: Thanks!

ha/monitor/k8s/object_monitor.py
Madhura-08: Thanks!

ha/monitor/k8s/object_monitor.py
Madhura-08: Thanks!

Madhura-08 (approved):

akash2144 (approved):

PR #676, opened 2022-03-24
CORTX-28313 : Add debug option to HA_start

mukhtar-inamdar:
Signed-off-by: Mukhtar Inamdar <mukhtar.p.inamdar@seagate.com>

# Problem Statement
- Add debug option to HA_start

# Design
-  Add debug option to HA_start (-d/--debug)
-  updated and corrected usage
- added flushing to ha_start

# Coding

# Testing 
[test_ha_start_debug_option_3.txt](https://github.com/Seagate/cortx-ha/files/8371749/test_ha_start_debug_option_3.txt)

# Review Checklist
- [x] PR is self reviewed
- [x] Is there a change in filename/package/module or signature? [Y/N]: **N**
- [x] If yes for above point, is a notification sent to all other cortx components? [Y/N] **N**
- [x] Side effects on other features (deployment/upgrade)? [Y/N] **N**
- [x] Dependencies on other component(s)? [Y/N] **N**

# Review Checklist 

# Documentation
  Checklist for Author
- [x] Changes done to WIKI / Confluence page / Quick Start Guide
[Debug option for cortx-ha](https://seagate-systems.atlassian.net/wiki/spaces/PUB/pages/956891431/Debug+option+for+cortx-ha)

### DIFF
--- a/ha/k8s_setup/ha_start.py
+++ b/ha/k8s_setup/ha_start.py
@@ -50,6 +50,15 @@ def start_driver_process():
     driver_process = Popen(['/usr/bin/python3', service_entry_mapping[args.services]], shell=False, stdout=PIPE, stderr=STDOUT) # nosec
     print(f"The driver process with pid {driver_process.pid}, and args {driver_process.args} started successfully.", flush=True)
 
+def is_debug_build():
+    # TODO: find out if it is debug build
+    return False
+
+def handle_debug_option():
+    if((args.debug&AUTO_RESTART)!=0):
+        print(f'Warning: Driver \'{args.services}\' exited with return code: {driver_process.poll()}, restarting again.', flush=True)
+        start_driver_process()
+
 def main(argv: list):
     try:
         print(str(args))
@@ -74,17 +83,15 @@ def main(argv: list):
         # Poll process.stdout to show stdout live
         while True:
             output = driver_process.stdout.readline()
-            if driver_process.poll() is not None:
-                if((args.debug&AUTO_RESTART)!=0):
-                    if output:
-                        print(output.strip().decode("utf-8"), flush=True)
-                    print(f'Warning: Driver \'{args.services}\' exited with return code: {driver_process.poll()}, restarting again.', flush=True)
-                    start_driver_process()
-
-                else:
-                    break
             if output:
                 print(output.strip().decode("utf-8"), flush=True)
+
+            if driver_process.poll() is not None:
+                if is_debug_build():
+                    handle_debug_option()
+                else:
+                    break
+
         exit(driver_process.poll())
     except Exception as proc_err:
         print(f'Driver execution stopped because of some reason: {proc_err}', file=sys.stderr, flush=True)
--- a/ha/k8s_setup/ha_start.py
+++ b/ha/k8s_setup/ha_start.py
@@ -5,6 +5,8 @@ import signal
 from subprocess import Popen, PIPE, STDOUT # nosec
 import sys
 
+AUTO_RESTART = 1
+
 driver_process = None
 
 my_parser = argparse.ArgumentParser(prog="ha_start",
@@ -13,7 +15,8 @@ my_parser = argparse.ArgumentParser(prog="ha_start",
                                     )
 
 my_parser.add_argument('--services', '-s', help='service_name', required=True, action='store')
-my_parser.add_argument('--config', '-c', help='config file', required=False, action='store')
+my_parser.add_argument('--config', '-c', help='config file url', required=False, action='store')
+my_parser.add_argument('--debug', '-d', help='debug level {0|1}', type=int, nargs='?', const=1, default=0, required=False, action='store')
 args = my_parser.parse_args()
 
 service_entry_mapping = {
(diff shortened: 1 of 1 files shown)


ha/k8s_setup/ha_start.py
```diff
@@ -13,7 +15,8 @@
                                     )
 
 my_parser.add_argument('--services', '-s', help='service_name', required=True, action='store')
-my_parser.add_argument('--config', '-c', help='config file', required=False, action='store')
+my_parser.add_argument('--config', '-c', help='config file url', required=False, action='store')
+my_parser.add_argument('--debug', '-d', help='debug level {0|1}', type=int, nargs='?', const=1, default=0, required=False, action='store')
```
Madhura-08, line 19: So, just for my understanding: 
If script is executed with:
-d/--debug and no value: debug value will be 1
no debug argument at all: debug value will be 0
And if given, that value will be used!
So, why not to give 1 for both const and default? Any reason to give 1 and 0 respectively for const and default?

Madhura-08, line 19: Why need to give debug option here? Don't you think AUTO_RESTART will be sufficient?
Or One point here, that restart should be configurable. In production, we can not afford restart again. But, then only debug is sufficient. Why need both debug and AUTO_RESTART?

ha/k8s_setup/ha_start.py
```diff
@@ -26,44 +29,65 @@ def usage(prog: str):
...
+        signal.signal(signal.SIGSEGV, handle_signal)
+        signal.signal(signal.SIGTERM, handle_signal)
+
+        start_driver_process()
+
+        # Poll process.stdout to show stdout live
+        while True:
+            output = driver_process.stdout.readline()
+            if driver_process.poll() is not None:
+                if((args.debug&AUTO_RESTART)!=0):
```
Madhura-08, line 78: I think no need to check != 0 here. & operation will result in 1 or 0. So, if 1 or if 0 can work here.

ha/k8s_setup/ha_start.py
```diff
@@ -26,44 +29,65 @@ def usage(prog: str):
...
+            f"[-d --debug   [level]] (if level is not provided, then it considered as 1)"
+            f"\n\t\t0 : No debug.\n\t\t1 : Auto restarts child process if it dia.", file=sys.stderr, flush=True)
 
 def handle_signal(signum, frame):
     """
     Signal handler call beck function, to forward the received signal to driver process
     """
-    sys.stdout.write(f"Signal {signum} at frame {frame} received.")
-    sys.stdout.write(f"Sending {signum} to driver process pid: {driver_process.pid}.")
+    print(f"Signal {signum} at frame {frame} received.", flush=True)
```
Madhura-08, line 44: Any concrete reason to use flush all over?

ha/k8s_setup/ha_start.py
mukhtar-inamdar: if i give 1 for default then debug value will be always 1 even if you don't want it to be.
I mean if you don't provide option as '-d' or '--debug' still it will be set to 1.
please check attached test file or https://seagate-systems.atlassian.net/wiki/spaces/PUB/pages/956891431/Debug+option+for+cortx-ha you will understand why the default value should be 0.

mukhtar-inamdar: the purpose is 'debug' not auto restart, auto restart is the solution so debug can make easy.
so if you provide --debug it will enable the auto-restart of the driver process so we can change/update the code easily and test it.
for more details you can read it here: https://seagate-systems.atlassian.net/wiki/spaces/PUB/pages/956891431/Debug+option+for+cortx-ha
So auto-restart is not an option it is the internal solution to enable change/update code and test it without restarting pod and no need to commit docker image.

And here I have given support for debug level 1 only tomorrow debug level 2 can do some addition work like enabling debug logging and more.

ha/k8s_setup/ha_start.py
mukhtar-inamdar: here I am performing a bitwise operation.
0x01 is auto_restart
0x10 is enable debug logging (in future)
0x11 is both (in future)
Here no matter which value is args.debug have I am checking if it has last bin enable i.e. auto restart

ha/k8s_setup/ha_start.py
mukhtar-inamdar: I restarted the health monitor 2 3 times see no log at all, so searched and used sys.stdout.flush() in between but it doesn't work.
so if there is overly less log then for a long time you will not able to see any log so this will not give us an error at the instance in any in the log of driver process. hence I need to add a flush option to all. the setting flush option is the only solution for me

ArchanaLimaye:
Please address review comments on the document "[Debug option for cortx-ha](https://seagate-systems.atlassian.net/wiki/spaces/PUB/pages/956891431/Debug+option+for+cortx-ha)"

ha/k8s_setup/ha_start.py
ArchanaLimaye, line 19: This and the remaining code should not be in mainstream.  You can have it under some option (that ensures that this is created only for dev build) or add TODO here for the same and file a ticket for that TODO

ha/k8s_setup/ha_start.py
```diff
@@ -26,44 +29,65 @@ def usage(prog: str):
...
+        signal.signal(signal.SIGINT, handle_signal)
+        signal.signal(signal.SIGSEGV, handle_signal)
+        signal.signal(signal.SIGTERM, handle_signal)
+
+        start_driver_process()
+
+        # Poll process.stdout to show stdout live
+        while True:
+            output = driver_process.stdout.readline()
+            if driver_process.poll() is not None:
```
gauri-bhosale, line 77: in below removed code this if condition breaking while loop but here you are continuing loop and break in else part, can u please check once again.

ha/k8s_setup/ha_start.py
Madhura-08: So, I think then the default will always be 0. Otherwise, if I execute with -d or --debug (without giving value), debug will be True which I may not want.

ha/k8s_setup/ha_start.py
Madhura-08: Ok..

ha/k8s_setup/ha_start.py
Madhura-08: Right. But, let's see combinations!
auto_restart is always 1. debug can be 1 or 0.
1 & 0: 0 (No auto restart, no debug)
1 & 1: 1 (Debug and auto restart)
Anything I am missing?

ha/k8s_setup/ha_start.py
mukhtar-inamdar: E.g.
ha_hart -s <service_name> -c <config_url>
ha_hart -s <service_name> -c <config_url> -d 0
ha_hart -s <service_name> -c <config_url> --debug 0
all above same and here debug level will be zero by default and even if you provide it specifically.
ha_hart -s <service_name> -c <config_url> -d
ha_hart -s <service_name> -c <config_url> --debug
ha_hart -s <service_name> -c <config_url> -d 1
ha_hart -s <service_name> -c <config_url> --debug 1
In this all case debug level is set to 1

ha/k8s_setup/ha_start.py
mukhtar-inamdar: (in future)
10 & 1: 0 enable debug log but no restart
11 & 1: 1 this value told us to enable auto restart no matter what other things are also required like enabling debug log or more options can be here. like (101 & 1: 1 and 110 & 1: 0 and 100 & 1: 0)

ha/k8s_setup/ha_start.py
Madhura-08: I got this. But if someone used just -d mistakenly (debug or auto restart is not intended). Then also, debug will be 1. But, as Archana mentioned, this we can make it available separately as part of dev build. So, its not the problem then. 
Fine for now. Thanks!

mukhtar-inamdar: ok , I got your point my intetion putting 1 as const or default with -d option is no other level are in use so -d means will help to debug and caller may not to look into this level thing.
But your point is valid will keep a note and first will understand arachana's point as I need to do for that thing.

ha/k8s_setup/ha_start.py
mukhtar-inamdar: ys because in case the auto-restart option is set by --debug flag will let continue to restart the process no matter why the child process was exited, killed, exit with an unhandled exception/crash.
so this is the aim so no need to break in this case.

gauri-bhosale: got it thanks.

gauri-bhosale (approved):

ha/k8s_setup/ha_start.py
Madhura-08: Discussed with Mukhtar. Cleared the confusion. Thanks!

mariyappanp (approved):

ha/k8s_setup/ha_start.py
mukhtar-inamdar: done

Madhura-08 (approved):

akash2144 (approved):

ha/k8s_setup/ha_start.py
ArchanaLimaye, line 19: Can this argument also be added conditionally ? i.e only when buld is debug build

mukhtar-inamdar: done

ArchanaLimaye (approved):

ArchanaLimaye:
@mukhtar-inamdar  please confirm that regular RPM (NON debug) builds fine and works fine.

gargpalak048 (approved):

PR #680, opened 2022-04-05
CORTX-29399 : When pod is restarted via replicas or kubectl delete, pod_restart count is not set properly.

gauri-bhosale:
Signed-off-by: gauri.bhosale <gauri.bhosale@seagate.com>

# Problem Statement
- When pod is restarted via replicas or kubectl delete, pod_restart count is not set properly.
- 

# Design
- system_health status checked incoming genericId and stored genericId if its not same 
checked for status but status not only my online it may 'starting' also  
# Coding
-  [X] Coding conventions are followed and code is consistent

# Testing 
- [X] Unit and System Tests are added
- [X] Test Cases cover Happy Path, Non-Happy Path and Scalability
- [X] Testing was performed with RPM
- https://jts.seagate.com/secure/attachment/513156/Test_Results.txt

# Review Checklist 
- [X] PR is self reviewed
- [X] JIRA number/GitHub Issue added to PR
- [X] Jira and state/status is updated and JIRA is updated with PR link
- [X] Check if the description is clear and explained

# Review Checklist 
- [X] Is perfline test run and the report with and without the changes updated in the PR? [Y/N]: 

Testing result :https://jts.seagate.com/secure/attachment/513156/Test_Results.txt

### DIFF
--- a/ha/core/system_health/system_health.py
+++ b/ha/core/system_health/system_health.py
@@ -420,9 +420,9 @@ class SystemHealth(Subscriber):
                     pod_restart_val = current_health_dict["events"][0]["specific_info"]["pod_restart"]
                     # Update the current health value itself.
                     latest_health = EntityHealth.read(current_health)
-                    if stored_genration_id and (stored_genration_id != incoming_generation_id):
-                        # TODO: Add stored_status != offline. 
-                        if (incoming_health_status == HEALTH_EVENTS.ONLINE.value) and (stored_status != HEALTH_EVENTS.FAILED.value):
+                    # TODO: Add stored_status != offline.
+                    if stored_genration_id and (stored_genration_id != incoming_generation_id) and (stored_status != HEALTH_EVENTS.FAILED.value):
+                        if (incoming_health_status == HEALTH_EVENTS.ONLINE.value):
                             # In delete scenario, online event comes first, followed by failed event.
                             # System health is expected to update the failed event first, then online event.
                             # If incoming is online event, change the stored event type to failed.
--- a/ha/core/system_health/system_health.py
+++ b/ha/core/system_health/system_health.py
@@ -421,6 +421,7 @@ class SystemHealth(Subscriber):
                     # Update the current health value itself.
                     latest_health = EntityHealth.read(current_health)
                     if stored_genration_id and (stored_genration_id != incoming_generation_id):
+                        # TODO: Add stored_status != offline. 
                         if (incoming_health_status == HEALTH_EVENTS.ONLINE.value) and (stored_status != HEALTH_EVENTS.FAILED.value):
                             # In delete scenario, online event comes first, followed by failed event.
                             # System health is expected to update the failed event first, then online event.
--- a/ha/core/system_health/system_health.py
+++ b/ha/core/system_health/system_health.py
@@ -426,6 +426,7 @@ class SystemHealth(Subscriber):
                             # System health is expected to update the failed event first, then online event.
                             # If incoming is online event, change the stored event type to failed.
                             # Update the failed event in system health and followed by incoming online event.
+                            # If stored_status is set as failed, then no need to send alert, as its replicaset pod down.
                             healthevent.specific_info = {"generation_id": stored_genration_id, "pod_restart": 1}
                             healthevent.event_type = "failed"
                             updated_health = SystemHealth.create_updated_event_object(healthevent.timestamp, current_timestamp, healthevent.event_type, healthevent.specific_info, latest_health)
--- a/ha/core/system_health/system_health.py
+++ b/ha/core/system_health/system_health.py
@@ -420,7 +420,7 @@ class SystemHealth(Subscriber):
                     # Update the current health value itself.
                     latest_health = EntityHealth.read(current_health)
                     if stored_genration_id and (stored_genration_id != incoming_generation_id):
-                        if incoming_health_status == HEALTH_EVENTS.ONLINE.value:
+                        if incoming_health_status in [HEALTH_EVENTS.ONLINE.value, status]:
                             # In delete scenario, online event comes first, followed by failed event.
                             # System health is expected to update the failed event first, then online event.
                             # If incoming is online event, change the stored event type to failed.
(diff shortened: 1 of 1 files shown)


ha/core/system_health/system_health.py
```diff
@@ -420,7 +420,7 @@ def process_event(self, healthevent: HealthEvent):
                     # Update the current health value itself.
                     latest_health = EntityHealth.read(current_health)
                     if stored_genration_id and (stored_genration_id != incoming_generation_id):
-                        if incoming_health_status == HEALTH_EVENTS.ONLINE.value:
+                        if incoming_health_status in [HEALTH_EVENTS.ONLINE.value, status]:
```
ArchanaLimaye, line 423: This will always return true since incoming_health_status and status both are mapping to status in healthevent

Madhura-08: I think status maps to already stored status.

mariyappanp: Possibility of status here can be 'starting' or 'online'. So considering both will be fine.
I agree with this change.

mukhtar-inamdar (approved):

mariyappanp (approved):

ha/core/system_health/system_health.py
Madhura-08: But, consider replicaset down case. 
1. When replicaset down, node status will be failed in system health.
2. Now, replicaset up, there is change in generation id and also incoming node status is online. 
So, this if will be passed and pod_restart count will be changed. But, in the replicaset scenario, we don't want that..!!

mukhtar-inamdar: once you did the required changes also write the comment for cases in which the restart count needs to be set/updated.
So this will make easy for someone to understand

ha/core/system_health/system_health.py
```diff
@@ -414,17 +414,19 @@ def process_event(self, healthevent: HealthEvent):
...
                     # Update the current health value itself.
                     latest_health = EntityHealth.read(current_health)
                     if stored_genration_id and (stored_genration_id != incoming_generation_id):
-                        if incoming_health_status == HEALTH_EVENTS.ONLINE.value:
+                        if (incoming_health_status == HEALTH_EVENTS.ONLINE.value) and (stored_status != HEALTH_EVENTS.FAILED.value):
                             # In delete scenario, online event comes first, followed by failed event.
                             # System health is expected to update the failed event first, then online event.
                             # If incoming is online event, change the stored event type to failed.
                             # Update the failed event in system health and followed by incoming online event.
+                            # If stored_status is set as failed, then no need to send alert, as its replicaset pod down.
```
mariyappanp, line 429: What if "stored status is offline"(Status map be supported in future)? Can you please check this.

gauri-bhosale: Currently Offline state is not in HEALTH_EVENT constants class, added Dod for offline in ticket : https://jts.seagate.com/browse/CORTX-30061

ha/core/system_health/system_health.py
gauri-bhosale: please review new change.

ha/core/system_health/system_health.py
mariyappanp: Okay. Thanks.

ha/core/system_health/system_health.py
```diff
@@ -414,17 +414,19 @@ def process_event(self, healthevent: HealthEvent):
...
                     stored_genration_id = current_health_dict["events"][0]["specific_info"]["generation_id"]
+                    stored_status = current_health_dict["events"][0]["status"]
                     incoming_generation_id = healthevent.specific_info["generation_id"]
                     incoming_health_status = healthevent.event_type
                     pod_restart_val = current_health_dict["events"][0]["specific_info"]["pod_restart"]
                     # Update the current health value itself.
                     latest_health = EntityHealth.read(current_health)
                     if stored_genration_id and (stored_genration_id != incoming_generation_id):
-                        if incoming_health_status == HEALTH_EVENTS.ONLINE.value:
+                        if (incoming_health_status == HEALTH_EVENTS.ONLINE.value) and (stored_status != HEALTH_EVENTS.FAILED.value):
```
Madhura-08, line 424: This will solve the problem when pod is deleted using replicaset.
But what if pod is deleted using kubectl delete where old pod statsus is starting or offline?

gauri-bhosale: Offline constant value currently not in HEATLH_EVENT class so for offline condition will add during ticket : https://jts.seagate.com/browse/CORTX-30061

ArchanaLimaye: @Madhura-08  if pod is "starting" it means that its "online" from monitor perspective; so we will need to check for restart in that case.
For "offline" that state will get added in CORTX-30061 and a TODO needs to be added here.
Also the DoD of CORTX-30061 needs to be updated to mention that this check here needs to be enhanced
@gauri-bhosale  as discussed already hope you have updated the DoD and adding a TODO here

Madhura-08: If required you can add that in the constant. But, I think its not needed. 
For replicaset, I gone though the code. But, I don't see any path where online event will be captured.
1. First if is to check generation_id is different or not -> Yes in case of replicaset
2. Check added by Gauri -> False
3. elif pod_restart_val -> False
After, no execution. 
@gauri-bhosale Can you please recheck?
cc @ArchanaLimaye

ArchanaLimaye: Yes @Madhura-08 . Either this check needs to be on line 423 OR an additional else is needed for this .
Gauri, please  address this and also the TODO

gauri-bhosale: yes @Madhura-08  above steps are correct.

gauri-bhosale: Done, and tested well,  please check new changes.

Madhura-08: Ok..This looks good now. Thanks!

Madhura-08 (approved):

mukhtar-inamdar (approved):

akash2144 (approved):

gargpalak048 (approved):
LGTM

ArchanaLimaye (approved):

ha/core/system_health/system_health.py
mukhtar-inamdar: I cannot resolve my comment alone if I resolve the whole conversation will get resolved,.
I am OK with changes let @ArchanaLimaye  resolve this conversion if she is OK as she is started this conversation.
CC @ArchanaLimaye

PR #712, opened 2022-05-09
CORTX-29667 : enable transaction push of kv consul for bulk push (performance change)

mukhtar-inamdar:
# Problem Statement
- ha_setup takes a long time for every consul put() call during mini provisioning.

# Design
-  enable transaction push of kv consul for bulk push (performance change)
-  implement caching for put call calls and put them all at a once

# Coding
-  [x] Coding conventions are followed and code is consistent

# Testing 
- [x] Unit and System Tests are added
- [x] Test Cases cover Happy Path, Non-Happy Path and Scalability
- [x] Testing was performed with RPM

# Review Checklist 
- [x] PR is self reviewed
- [x] JIRA number/GitHub Issue added to PR
- [x] Jira and state/status is updated and JIRA is updated with the PR link
- [x] Check if the description is clear and explained
- [x] Is there a change in filename/package/module or signature? [Y/N]:  N
- [x] If yes for above point, is a notification sent to all other cortx components? [Y/N] NA
- [x] Side effects on other features (deployment/upgrade)? [Y/N] N
- [x] Dependencies on other component(s)? [Y/N] N
-     If yes for above point, post link to the corresponding PR.
- Testing: https://jts.seagate.com/secure/attachment/517504/CORTX-29667.log.v3.txt
Note: debug logs are coneveted to INFO for testing but it in code these all are debug.

# Review Checklist 
- [ ] Is perfline test run and the report with and without the changes updated in the PR? [Y/N]: 

# Documentation
  Checklist for Author
- [ ] Changes done to WIKI / Confluence page / Quick Start Guide

PR #713, opened 2022-05-11
CORTX-29667 : Added caching in ConsulKvStore for enabeling batch put

mukhtar-inamdar:
Signed-off-by: Mukhtar Inamdar <mukhtar.p.inamdar@seagate.com>

replicate the pull request: https://github.com/Seagate/cortx-ha/pull/712
All comments from #712: https://jts.seagate.com/secure/attachment/518021/CORTX-29667-comments-pull-%23712.pdf
(JSON): https://jts.seagate.com/secure/attachment/518023/CORTX-29667-comments.json

# Problem Statement
- ha_setup takes a long time for every consul put() call during mini provisioning.

# Design
-  enable transaction push of kv consul for bulk push (performance change)
-  implement caching for put call calls and put them all at a once

# Coding

# Testing 
- [x] Test Cases cover Happy Path, Non-Happy Path and Scalability

# Review Checklist 
- [x] PR is self reviewed
- [x] Jira and state/status is updated and JIRA is updated with the PR link
- [x] Is there a change in filename/package/module or signature? [Y/N]:  N
- [x] If yes for above point, is a notification sent to all other cortx components? [Y/N] NA
- [x] Side effects on other features (deployment/upgrade)? [Y/N] N
- [x] Dependencies on other component(s)? [Y/N] N
- Testing: https://jts.seagate.com/secure/attachment/517504/CORTX-29667.log.v3.txt
Note: debug logs are converted to INFO for testing but in code, these all debug.

# Review Checklist 

# Documentation
  Checklist for Author

### DIFF
--- a/ha/k8s_setup/ha_setup.py
+++ b/ha/k8s_setup/ha_setup.py
@@ -261,7 +261,10 @@ class ConfigCmd(Cmd):
             # TODO: Verify whether these newly added config is avilable in the confstore or not
             with open(const.HA_CONFIG_FILE, 'w+') as conf_file:
                 yaml.dump(conf_file_dict, conf_file, default_flow_style=False)
-            self._confstore = ConfigManager.get_confstore()
+            # Note: kv_enable_batch is setting True to commit all the kv keys in 1 time for time optimization.
+            # if kv_enable_batch set to true then commit operation needs to run to put all the key values in store.
+            kv_enable_batch_put = True
+            self._confstore = ConfigManager.get_confstore(kv_enable_batch=kv_enable_batch_put)
 
             Log.info(f'Populating the ha config file with consul_endpoint: {consul_endpoint}')
 
@@ -291,6 +294,11 @@ class ConfigCmd(Cmd):
             # Stopped disk, cvg resource key addition to consul to reduce consul accesses
             # till CORTX-29667 gets resolved
             #self._add_cvg_and_disk_health()
+
+            # Note: if batch put is enabled needs to commit
+            # to push all the local cashed values to consul server
+            if kv_enable_batch_put:
+                self._confstore.commit()
 
             Log.info("config command is successful")
             sys.stdout.write("config command is successful.\n")
--- a/ha/util/consul_kv_store.py
+++ b/ha/util/consul_kv_store.py
@@ -15,9 +15,19 @@
 # about this software or licensing, please email opensource@seagate.com or
 # cortx-questions@seagate.com.
 
+from base64 import b64encode
 import consul
+from consul import ConsulException
+from consul.base import ClientError
+from requests.exceptions import RequestException
+from urllib3.exceptions import HTTPError
 import socket
 from ha.const import HA_DELIM
+from cortx.utils.log import Log
+from typing import Any, Dict, List, NamedTuple, Optional
+
+TxPutKV = NamedTuple('TxPutKV', [('key', str), ('value', str),
+                                 ('cas', Optional[Any])])
 
 #TODO: Update set/get/update function to provide blocking and non blocking function
 class ConsulKvStore:
@@ -25,7 +35,7 @@ class ConsulKvStore:
 
     _keys = {}
 
-    def __init__(self, prefix: str, host: str="localhost", port: int=8500):
+    def __init__(self, prefix: str, host: str="localhost", port: int=8500, enable_batch: bool=False):
         """
         Consul KV store.
 
@@ -43,6 +53,9 @@ class ConsulKvStore:
         self._verify(prefix, host, port)
         self._consul = self._get_connection(prefix, host, port)
         self._consul.kv.put(self._prepare_key(""), None)
+        self._enable_batch_put = enable_batch
+        if self._enable_batch_put:
+            self._payload = {}
 
     def _verify(self, prefix: str, host: str, port: int):
         """
@@ -103,10 +116,20 @@ class ConsulKvStore:
         Return:
             bool: True if key exists else False.
         """
-        _, data = self._consul.kv.get(self._prepare_key(key), recurse=True)
-        if data is None:
-            return False
-        return True
+        k = self._prepare_key(key)
+
+        # First find in cache if batch put is enabled
+        Log.debug(f"[consul-op] is key exist in cache: {k}")
+        if self._enable_batch_put and [key for key in self._payload if key.startswith(k)]:
+            Log.debug(f"[consul-op] key {k} found in cache: True")
+            return True
+
+        # find in consul and check if exists
+        Log.debug(f"[consul-op] is key exist: {k}")
+        _, data = self._consul.kv.get(k, recurse=True)
+        Log.debug(f"[consul-op] key {k} found in consul: True")
+
+        return True if data else False
 
     def set(self, key: str, val: str=None):
         """
@@ -122,7 +145,16 @@ class ConsulKvStore:
         self._verify_data(key)
         if self.key_exists(key):
             raise Exception(f"Key {key} already exists in kv store.")
-        self._consul.kv.put(self._prepare_key(key), val)
+        k = self._prepare_key(key)
+
+        if self._enable_batch_put:
+            Log.debug(f"[consul-op] Putting key value in cache for key: {k}")
+            self._payload[k] = val
+            return val
+
+        Log.debug(f"[consul-op] Putting key value: {k}")
+        self._consul.kv.put(k, val)
+        Log.debug(f"[consul-op] put Key value: {val}")
         return val
 
     def update(self, key: str, new_val: str):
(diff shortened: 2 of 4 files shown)


mssawant:
@mukhtar-inamdar, kindly reword the commit message as follows,
```
CORTX-29667 : HA configuration upload is in-efficient


HA maintains a part of Cortx cluster configuration in Consul KV store.
A set of relevant key-values are uploaded as part of configuration during
deployment. The present upload mechanism is in-efficient, as a single key
value is ulploaded per put request. Alternatively, as Consul supports batch
operations, several key-values can be uploaded in a single PUT transaction.

Solution:
- Use Consul transactional PUT operation to upload several HA key-values.

Signed-off-by: Mukhtar Inamdar <mukhtar.p.inamdar@seagate.com>
```

ha/k8s_setup/ha_setup.py
```diff
@@ -261,7 +261,10 @@ def process(self):
             # TODO: Verify whether these newly added config is avilable in the confstore or not
             with open(const.HA_CONFIG_FILE, 'w+') as conf_file:
                 yaml.dump(conf_file_dict, conf_file, default_flow_style=False)
-            self._confstore = ConfigManager.get_confstore()
+            # Note: kv_enable_batch is setting True to commit all the kv keys in 1 time for time optimization.
+            # if kv_enable_batch set to true then commit operation needs to run to put all the key values in store.
+            kv_enable_batch_put = True
```
mssawant, line 266: Why not have separate functions, `_confstore::commit()` and `_confstore::commit_batch()` anyway you are setting `kv_enable_batch_put` to `True` explicitly?

ha/util/consul_kv_store.py
```diff
@@ -122,7 +145,16 @@ def set(self, key: str, val: str=None):
         self._verify_data(key)
         if self.key_exists(key):
             raise Exception(f"Key {key} already exists in kv store.")
-        self._consul.kv.put(self._prepare_key(key), val)
+        k = self._prepare_key(key)
+
+        if self._enable_batch_put:
+            Log.debug(f"[consul-op] Putting key value in cache for key: {k}")
+            self._payload[k] = val
```
mssawant, line 152: This is not very good way to implement, functions need to be more explicit with respect to the parameters that they accept. Flags don't help much with the readability. Better to have an explicit function, e.g. `commit_batch()` that accepts a list of Key-values.

ha/util/consul_kv_store.py
```diff
@@ -122,7 +145,16 @@ def set(self, key: str, val: str=None):
         self._verify_data(key)
         if self.key_exists(key):
```
mssawant, line 146: How is this `key_exists` exception is expected to be handled by the user? Consul wise, if the kv already exists then it will be simply overwritten by new data. If you don't want to overwrite then you can instead do a CAS transaction with `cas=0`, with this, if key already exists, kv put operation will automatically fail and it will save a get request effort.

ha/util/consul_kv_store.py
```diff
@@ -154,13 +195,31 @@ def get(self, key: str = ""):
...
+
+        Log.debug(f"[consul-op] getting the value for key: {k}")
+        _, data = self._consul.kv.get(k, recurse=True)
+        Log.debug(f"[consul-op] got the value for Key: {k}")
+
         key_val: dict = {}
-        for key in data:
-            key_val[key['Key']] = key['Value'].decode("utf-8") if isinstance(key['Value'], bytes) else key['Value']
-        return key_val
+        if data:
```
mssawant, line 213: Why are we fetching all the keys with a given prefix and not the exact ones?

ha/k8s_setup/ha_setup.py
mukhtar-inamdar: if 'kv_enable_batch_put' you set to False you don't need to call commit() in fact the commit() is only for batch operation.
I have added check in commit function as it should be called only in batch operation.

ha/k8s_setup/ha_setup.py
```diff
@@ -261,7 +261,10 @@ def process(self):
...
+            self._confstore = ConfigManager.get_confstore(kv_enable_batch=kv_enable_batch_put)
```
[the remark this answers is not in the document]
mukhtar-inamdar: yes, it is used like protected, but it's an existing class member I have not added newly.
it is used in 3 classes.

I am thinking not to change in this PR as these will be cleanup/correction changes, unnecessary it will increase the code chnages of this PR, but yes need to change.

ha/util/consul_kv_store.py
mukhtar-inamdar: Yes, i.e. good same discussed with Neerav also,

but in ha_setup() we are not directly adding keys we are generating events that get proceeded by the event manager and during the process of each event it generates and adds required key(s), so at no point in time you will have a bunch of keys to add so I have taken an approach to catch is in kV store class without changing the code event manager otherwise it will be design change for the event manager and may need to think more.

Yes, we can write a separate function commit_batch() and put call the keys directly from ha_setup() and don't process events through the event manager.
But needs to understand and discuss if it is possible and what are the consequences of skipping event processing otherwise we land up adding too much code in ha_setup and maybe it is duplicating much node code.

Hi @mssawant @shaileshsvc, @ArchanaLimaye, @NeeravChoudhari,
Needs to discuss whether is above change is feasible now and its consequences. otherwise can schedule later.
Need input on this.

ha/util/consul_kv_store.py
mukhtar-inamdar: Hey, I checked all the calls caller calling key_exists() or get to check if it already exists.
So looks like we can remove this call and exception both.

ha/util/consul_kv_store.py
mukhtar-inamdar: consul.get() works like this when we give some key and recurse=True it finds out all the keys with prefix and I think this functionally is being used here I am added similar functionality that consul.kv.get(key,recurse=True) gives us.

ha/util/consul_kv_store.py
```diff
@@ -176,7 +235,27 @@ def delete(self, key: str = "", recurse: bool = False):
             str: Return dictionary of all key val pair.
         """
         data = self.get(key)
-        self._consul.kv.delete(self._prepare_key(key), recurse=recurse)
+        k = self._prepare_key(key)
+
+        # if enabled batch put then delete from payload if any new value exist
```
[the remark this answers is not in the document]
mukhtar-inamdar: 1. if recurse is True then there are many keys some are in payload and some are already in consul so needs to delete both.
2. in case if recurse is False the new value is in cache but there can be old value for same key is exist in consul needs to be deleted otherwise it will work like undo update and not delete.

ha/util/consul_kv_store.py
mssawant: @mukhtar-inamdar, See, anyway its the HA that is generating the key on receiving the event. Is it that HA doesn't know before hand what type of events it will receive?

ha/util/consul_kv_store.py
mssawant: Okay, do you expect lot of work for this to be done in the ongoing PR?

ha/util/consul_kv_store.py
mukhtar-inamdar: Hi @mssawant,
CC @ArchanaLimaye 
Below are the points from the discussion in the group:

Yes, actually we are initialing HA through events, so that can be skipped, but needs to understand the consequences.

So just add keys with default values and once the deployment is done, start processing upcoming events from the K8s monitor will look like work, but needs to evaluate that keys with the default value, can be just hardcoded or needs to generate by some code,

this may take the effort of story points 3 to 5. @ArchanaLimaye  your thoughts, please?

ha/k8s_setup/ha_setup.py
mssawant: Okay, so you mean there are some events which are saved in persistent store(Consul) and some not?
In any case, the way commit is invoked, it doesn't really seem to be related to `kv_enable_batch_put` or batching. I think commit can happen in a batch or not. It depends on the use case. That's why these things need to be explicit.
From this, I see that there can be multiple versions of commit, committing to a persistent store or not. This needs to be classified per event type and not really with respect to commit type.

mssawant (approved):
Approving the current version of PR as it solves the corresponding deployment failure problem. Suggested design changes to be taken up separately.

> The Separate Task created to track these changes: https://jts.seagate.com/browse/CORTX-31319

shaileshsvc (approved):

ArchanaLimaye (approved):

ha/k8s_setup/ha_setup.py
mukhtar-inamdar: will fix in: https://jts.seagate.com/browse/CORTX-31319

ha/k8s_setup/ha_setup.py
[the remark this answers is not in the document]
mukhtar-inamdar: will fix in: https://jts.seagate.com/browse/CORTX-31319

ha/util/consul_kv_store.py
mukhtar-inamdar: will fix in: https://jts.seagate.com/browse/CORTX-31319

ha/util/consul_kv_store.py
mukhtar-inamdar: will fix in: https://jts.seagate.com/browse/CORTX-31319

ha/util/consul_kv_store.py
mukhtar-inamdar: Addressed

ha/util/consul_kv_store.py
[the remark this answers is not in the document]
mukhtar-inamdar: Addressed

ArchanaLimaye:
Keeping the conversations open as some of them are going to be fixed in 
https://jts.seagate.com/browse/CORTX-31319 Design Change for Optimization in HA Mini Provisioning

PR #721, opened 2022-08-17
CORTX-33787: [v0.9.0][2.0.0-880] Kafka errors UNKNOWN_TOPIC_OR_PART d…

Madhura-08:
…uring build deployment
  
Problem:
HA mini provisioning is failing because of Kafka connection. It is happening because Kafka and HA pod is getting deployed simultaneously. For that, HA needs to try to reconnect/retry. But that is not happening. Hence Kafka topics consul keys and other keys are not getting created. So, HA POD is running but not functional.

Solution:
To reconnect/retry, the init container needs to be restarted because mini provisioning gets executed as part of the init container. For the init container to restart, a proper failure code must be returned to the caller. For here, the exception needs to be re-raised to the caller and there, already the error code returning is handled.

# Problem Statement
https://jts.seagate.com/browse/CORTX-33787

# Design
HA and third party kafka pods now gets deployed simultaneously. HA connects to kafka at its init stage(mini provisioning) for creating topics. As HA tries to connect to kafka, but that time, kafka was running but it is not ready to serve. Hence HA fails at mini provisioning stage and fails to create consul keys. For this init container needs to be restarted so that kafka connection retries will be executed. Ideally init container is meant to be executed only once. It will be restarted only if some failure occurs. And failure can be propagated in the form of return code. 
From HA side, the exception was not getting re-raised and that is why return code was always sent as 0 which was not causing the restart of init container. So, re-raising the exception and proper return code handling is needed.

# Coding

# Testing 
https://jts.seagate.com/secure/attachment/532015/CORTX-33787_test_results.txt

# Review Checklist 
- [x] PR is self reviewed

# Review Checklist 

# Documentation
  Checklist for Author

### DIFF
--- a/ha/k8s_setup/ha_setup.py
+++ b/ha/k8s_setup/ha_setup.py
@@ -337,15 +337,19 @@ class ConfigCmd(Cmd):
         except TypeError as type_err:
             Log.error(f'HA config command failed: Type mismatch: {type_err}.\n')
             sys.stderr.write(f'HA config command failed: Type mismatch: {type_err}.\n')
+            raise
         except yaml.YAMLError as exc:
             Log.error(f'HA config command failed: Invalid yaml configuration: {exc}.\n')
             sys.stderr.write(f'Ha config failed. Invalid yaml configuration: {exc}.\n')
+            raise
         except OSError as os_err:
             Log.error(f'HA config command failed: OS_error: {os_err}.\n')
             sys.stderr.write(f'HA Config failed. OS_error: {os_err}.\n')
+            raise
         except Exception as c_err:
             Log.error(f'HA config command failed: {c_err}.\n')
             sys.stderr.write(f'HA config command failed: {c_err}.\n')
+            raise
 
     def _add_cluster_component_health(self) -> None:
         """


mssawant:
@Madhura-08, following rules are exercised in Hare commit messages, 
```
CORTX-33787: [v0.9.0][2.0.0-880] Kafka errors UNKNOWN_TOPIC_OR_PART during build deployment

- re-raise the exception in order to properly propagate the script return code
  to caller

```
1. keep the summary line short (80 cols)
2. Its good to describe the problem a bit. From the commit message I am not able to understand what the problem was and why are we implementing the fix.
3. Its good to separate classify the Solution with a Solution tag.

Please consider a following commit message format,
```
CORTX-33787: ha deployment fails due to kafka errors

<Describe the problem, e.g. unknown topic exception not handled>

Solution:
<Describe the solution, mainly how solution fixes the problem>

```

ha/k8s_setup/ha_setup.py
```diff
@@ -337,15 +337,19 @@ def process(self):
         except TypeError as type_err:
             Log.error(f'HA config command failed: Type mismatch: {type_err}.\n')
             sys.stderr.write(f'HA config command failed: Type mismatch: {type_err}.\n')
+            raise
```
mssawant, line 340: instead of re-raising, cannot we return an error code?

Madhura-08: This exception gets caught by the caller:
https://github.com/Seagate/cortx-ha/pull/721/files#diff-3961fe8f57cf6e98896333df1e8846fb536f6a2b6384bc51019b15822a432913R587
And there, the return code is returned. Here it was not raised and that is why there was a problem. So, I fixed it by re-raising.
The return code will definitely be going to return from now.

Madhura-08 (commented):
@mssawant  Sorry, I missed following the commit message format. I fixed that now. Please take a look. Thanks!

ha/k8s_setup/ha_setup.py
mssawant: Okay.

mssawant (approved):

